PR #6, opened 2022-06-20
fix: fix truncated log messages

teqm:


gquintard:
good catch! the bug is still a bit confusing because we shouldn't need the `\0` as we specify the length, I'll check if there's something to do in `varnishtest`. Thanks for your contribution!

Issue #15, opened 2023-03-05
Request for mutability of Ctx for VDPs and VFPs 


dward:
I would like to request for the mutability of Ctx for both the VDP and VFP new() method. Currently, it is impossible to modify the response headers, such as removing the Content-Length header or setting Content-Encoding, without using unsafe calls. I'm somewhat new to rust, so feel free to call me out if I'm wrong here.

Furthermore, to prevent varnish from responding with incorrect content-length, there is a need to set ctx.raw.req.resp_len to -1, which also requires an unsafe call. Therefore, I am requesting a safe way to set the resp_len.

It would also be beneficial to provide a safe interface to the RFC2616_Weaken_Etag method as well. I can create separate issues for this if needed.

Thank you for your hard work on this project!

gquintard:
Hi @dward!

Thanks for your interest, it's great to see people using this code base!

Let's take care of the quick one first: yes, please open a new issue for `RFC2616_Weaken_Etag` (we might just replicate the logic using `rust` by the way, since the code is so small, but let's discuss it elsewhere)

Regarding, `VDP/VFP::new()` method, it would make sense to have a `mut Ctx` and I've very open to it (specially if you are willling to open a PR ;-) ), but I'm thinking this could be an opportunity to make the whole interface better (like, not forcing the implementer to add a `\0` at the end of the VDP/VFP name...)

May I ask what your VDP/VFP do to get a better idea of where to focus our efforts?

gquintard:
for context, here what one of my own vmod does: https://github.com/gquintard/vmod_rers/blob/main/src/lib.rs#L230-L240

i.e. it accesses the raw `ctx` pointer to modify the `content-length` header, at the time I didn't want to over-generalize and was happy with calling the `C` functions directly, but as more use-cases emerge, we need to make this easier and safe

dward:
@gquintard - Thanks for the quick response! 

I'm currently experimenting with writing a VFP for Brotli compression (so that varnish will store the cache as Brotli compressed), and a VDP for Brotli decompression (for clients that do not support Brotli), and an additional VDP for gzip compression (for clients that only support gzip).

I'll open a PR too with the new() mutability changes, as I already patched it internally. 

Regarding the content-length from a vdp, I found it necessary to set ctx.raw.req.resp_len to -1. Initially I had just unset the content-length, but I found that the original content length was still being served. I was calling:

```
        if let Some(ref mut resp) = mut_ctx.http_resp {
[...]

PR #16, opened 2023-03-06
Ctx should be mutable in the new method of VDPs and VFPs

dward:
This resolves issue #15

It changes the mutability of Ctx in the new method in both the VDP/VFP interfaces to allow modifications of things such as headers.


gquintard:
thank you Dustin. I'll update https://github.com/gquintard/vmod_rers to make use of that

Issue #20, opened 2023-05-15
wrap_vfp_fini needs null guard around drop


dward:
wrap_vfp_fini appears to need a null check around the drop of priv1 as it may remain null during a failure or an explicit pass during init. The drop method panics when encountering a null pointer.

I'm able to reproduce the crash by modifying examples/vmod_vfp/src/lib.rs to return InitResult::Pass from the new method:

```
fn new(_: &mut Ctx, _: &mut VFPCtx) -> InitResult<Self> {
    InitResult::Pass
}
```

and by modifying the struct to contain alloc'd data:
```
struct Lower {
    buffer: Vec<u8>,
}
```

gquintard:
Good catch, thank you Dustin. Can ask you for a VTC in the PR?
You might need to create a new function in the test vmods, but that's
definitely a check worth having.

dward:
@gquintard - Sure thing. I'll update the PR once I create a VTC.

gquintard:
Thanks! Just for information, I'm away from computers until may 27th, so
expect lag, but I'm definitely merging this as soon as I can

dward:
@gquintard - Sounds good! I've updated the pull with the test.

gquintard:
fixed by #21

PR #21, opened 2023-05-15
add null check to wrap_vfp_fini

dward:
Fixes issue #20


gquintard:
Thanks @dward , merging this.

Side note, [this PR](https://github.com/varnishcache/varnish-cache/pull/3772) will probably become important after the Varnish September release as it will make the VFP/VDP dance simpler, but that will probably also change the ergonomics a bit. Would you like be to reach out when I have something, since I know you'll probably be impacted?

dward:
@gquintard Thanks for the heads up! That would be great.

Issue #25, opened 2023-08-13
devcontainer definition


bugfest:
A [devcontainer](https://containers.dev) definition can help developers to bootstrap their dev environments. I think `varnish-rs` is a perfect fit for this as it depends on specific software versions (`python >= 3`, `varnish >= 7`). 

This will also help to deal with specific installations like the one defined in the `tests` gh action: https://github.com/gquintard/varnish-rs/blob/ed56f08ccc9daf57683a096b505e1f1709cb9fbd/.github/workflows/tests.yaml#L22-L24

bugfest:
Hi @gquintard, thanks for the reply. I totally agree with this. I'll check if https://github.com/devcontainers/ci/blob/main/docs/github-action.md would be a good fit.

PR #26, opened 2023-08-13
devcontainer definition

bugfest:
Implements #25


gquintard:
it's been a year, timing out

PR #28, opened 2024-08-30
fix a few minor lints

nyurik:
Used `cargo clippy` to fix a few minor complains. Also, hid a few warnings in the auto-genned code.

Note that `cargo test` from the root dir does not compile (should this be added to CI?)


gquintard:
Hi! Thanks for taking the time to look at this (and for all the PRs)

`cargo test` should definitely be part of CI, can I let you open a PR for it too?

I'll fix the compilation issue

PR #33, opened 2024-09-04
Improve github action workflow

nyurik:
P.S. Note that CI currently fails - not certain why yet

I copied some of these from many of my crates like https://github.com/nyurik/sqlite-hashes (eventually I think it is better to use `just` to simplify the process, and make it identical for the developer and CI)

* removed the `no-libvarnish` job -- not used
* remove rust installation - it is always pre-installed unless you want an older version, e.g. for msrv test
* use build cache
* add all sorts of tests
* no need to do a loop over vmod - enough to just say `cargo ... --workspace`
* not much value in `RUST_BACKTRACE=1` I think - if it fails, longer backtrace won't tell you much in CI
* add doc testing

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,10 +26,8 @@ jobs:
           sudo apt-get install -y varnish-dev
       - name: test-fmt
         run: cargo fmt --all -- --check
-      - name: check
-        run: RUSTFLAGS='-D warnings' cargo check --workspace --all-targets
-      - name: build vmod
-        run: cargo build --all-targets
+      - name: build
+        run: RUSTFLAGS='-D warnings' cargo build --workspace --all-targets
       - name: test
         run: cargo test --workspace --all-targets
       - name: test-doc
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -1,36 +1,50 @@
 name: Build
-on: [push]
-jobs:
-  no-libvarnish:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-      - run: |
-          cargo doc
-  with-libvarnish:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-      - run: |
-          set -x
-          sudo apt-get install -y curl
-          curl -s https://packagecloud.io/install/repositories/varnishcache/varnish75/script.deb.sh | sudo bash
-          sudo apt-get install varnish-dev
-      - run: |
-          set -x
-          cargo doc
-          export RUST_BACKTRACE=1
-          for dir in vmod_test $(ls -d examples/vmod_*); do
-            (
-              cd $dir
-              cargo build --all-targets
-              cargo test
-            )
-          done
 
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  release:
+    types: [ published ]
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      RUST_BACKTRACE: 1
+    steps:
+      - uses: actions/checkout@v4
+      - name: rust-info
+        run: rustc --version && cargo --version
+      - uses: Swatinem/rust-cache@v2
+        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+      - name: install varnish-dev
+        run: |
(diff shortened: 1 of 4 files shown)


gquintard:
don't worry about the CI failure, it did it's job and quote a bug :) I'll fix it (in short: we can be called with 0-length buffer, with a NULL pointer, and `std::slice::from_raw_parts_mut()` is rightfully unhappy with it)

re: `RUST_BACKTRACE=1`, I actually just added it, and it allowed be to debug the issue above, so I would keep it :-D

TIL `cargo ... --workspace` that's great, and much simpler

I'd like to keep `no-libvarnish`, it makes sure that crates.io doesn't blow up when creating the docs without `varnish` being installed

nyurik:
thx for the explanation, updated

.github/workflows/tests.yaml
```diff
@@ -1,36 +1,50 @@
 name: Build
-on: [push]
+
+on:
```
gquintard, line 3: any reason to limit ourselves here?

.github/workflows/tests.yaml
```diff
@@ -1,36 +1,50 @@
...
+      - name: install varnish-dev
+        run: |
+          curl -s https://packagecloud.io/install/repositories/varnishcache/varnish75/script.deb.sh | sudo bash
+          sudo apt-get install -y varnish-dev
+      - name: test-fmt
+        run: cargo fmt --all -- --check
+      - name: check
+        run: RUSTFLAGS='-D warnings' cargo check --workspace --all-targets
+      - name: build vmod
+        run: cargo build --all-targets
```
gquintard, line 32: you need to build all the workspace as the `cargo test` below will require the vmods to be built

.github/workflows/tests.yaml
nyurik: this is a common pattern for FOSS projects:
* only build the "release" branches (like `main`)
* when you or someone else pushes modification branches, there is little value in building those branches directly. Instead, you want to build the pull request **into** one of the release branches (main)
* this allows you and everyone else contributing to this project to work on the same system - me or you or anyone must create a PR for it to be built

.github/workflows/tests.yaml
nyurik: i was a bit confused by that - test (in theory) must build the code. How is it possible for the test to run without building it first?

.github/workflows/tests.yaml
gquintard: I mean, I have no objections, but we could also leave it as a free for all, it's not like we have millions of commits every day :-)

.github/workflows/tests.yaml
nyurik: hmm, i guess in theory `test` might skip the step in which build places final binaries to the `release|debug` dir... Ok, will have to simply modify check -> build.

.github/workflows/tests.yaml
nyurik: ??  it is free for all - my point is that you don't want to compile it 3 times:  on branch creation, on pull creation, and on merge to `main` -- simply because the first two are identical, and you should have PR before merging.

nyurik: P.S. note that if you modify the branch from which a PR was created, that triggers rebuild on both `branch` and `pull` -- so you get two builds on each change

gquintard: from my understanding, currently we run CI for every commit (i.e. free for all), and with this PR we'll only run for a subset of commits

The proposed way works for me, we're good

nyurik: not exactly - it runs when **you** commit anything to your repo.

* When I create a new PR, I first create a branch -- which trigger a local action run  (on my own fork). The results are local to my own fork.
* Afterwards, I create a PR, but you do NOT see the results of a build -- because the build happened on BRANCH before the PR was created (i.e. the build is not associated with the PR).
* Whenever I update the branch (e.g. per your feedback), the build once again happens locally on my fork - the PR does not show its results

So in short, it is far better to build on PR change rather than branch change - because it keeps all those build results in one place - associated with the PR

gquintard: ooooooh, gotcha, you're right, it's much better

gquintard:
ok to merge for me, shall I press the green button?

nyurik:
yep :)

nyurik:
btw, you may want to do "squash only merges" - keeps history cleaner

gquintard:
I prefer to just rebase, merging makes it harder to `git bisect`*, feel free to squash and rebase though. I'll merge later today to give you some time if you want to do that

* yes, the history is a mess a the moment, the pro/con of working alone on a code base, but hey, that's the intention that counts!

nyurik:
its already on top, feel free to merge whenever, i won't change

PR #34, opened 2024-09-04
Remove varnish-sys

nyurik:
Moved all buindings to `crate::raw` (initially i used `varnish_sys` as a submodule, but the full path is a bit weird - `varnish::varnish_sys::*`, so `raw` is shorter.

Also, I searched github for `include!(concat!(env!("OUT_DIR"), "/bindings.rs"));`, and I noticed `raw`, `ffi`, `bindings`, and project specific words like `gl` being used for the submodule name.

Lastly, I used `cargo +nightly fmt --all -- --config imports_granularity=Module,group_imports=StdExternalCrate` to reformat all the use statements - keeps them a bit cleaner.


gquintard:
I need to ruminate this one a bit (and conflicts need to be fixed) but it's a yes in principle

note: and I know I'm one to talk, but the PR would be easier to review if they were broken up in more or less atomic commits. We don't have a lot of traffic here, so I'm not going to be stringent about it, though.

nyurik:
I agree about atomic commits - but this one is the smallest possible PR for merging it:  you must do the renaming in one go.  In theory, i could have kept the name `varnish_sys`, in which case i would have had to just add `use crate::varnish_sys` at the top of all files, and then later we could have renamed it, but that's the only change that can be optimized this way.

gquintard:
not exactly true, there's a bunch of linting also happening all over the place. They don't have to be in the same commit.
Anyway, I don't want to sound ungrateful, I realy, really appreciate your work and attention here

nyurik:
oh, i thought i kept these to minimum, sorry

gquintard:
all good, no worries

nyurik:
I rebased and renamed it back to `varnish_sys` -- this way we can refactor it separatelly.

gquintard:
sweet, merging

PR #35, opened 2024-09-04
A few small lints

nyurik:
* reformat with `cargo +nightly fmt --all -- --config imports_granularity=Module,group_imports=StdExternalCrate`
* add clippy lint configuration to Cargo.toml

This seems like a bug, but I didn't change the logic:

```rust
        self.raw.protover = match value {
            "HTTP/0.9" => 9,
            // FIXME: Is this a bug?
            "HTTP/1.0" | "HTTP/1.1" => 10,
            "HTTP/2.0" => 20,
            _ => 0,
        };
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+lints.workspace = true
+
 [package]
 name = "varnish"
 version = "0.0.19"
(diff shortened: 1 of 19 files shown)


gquintard:
thanks, the `protover` mimicked what happened in the C code a while back, looks like it has changed and we can do the sane thing now. I'll change it

PR #36, opened 2024-09-04
More linting

nyurik:
Make all crates use the same linting rules, and fixed a few.

Note that I fixed `extra_unused_lifetimes` for `unsafe extern "C" fn chunk_collector(...)` -- which looked a bit suspicious, but seems to be ok.  Need CI tests! :)

### DIFF
--- a/src/vcl/ctx.rs
+++ b/src/vcl/ctx.rs
@@ -127,13 +127,13 @@ impl<'a> Ctx<'a> {
     }
 
     pub fn cached_req_body(&mut self) -> Result<Vec<&'a [u8]>, crate::vcl::Error> {
-        unsafe extern "C" fn chunk_collector<'a>(
+        unsafe extern "C" fn chunk_collector(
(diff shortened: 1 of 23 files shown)


src/vcl/ctx.rs
```diff
@@ -127,13 +127,13 @@ impl<'a> Ctx<'a> {
     }
 
     pub fn cached_req_body(&mut self) -> Result<Vec<&'a [u8]>, crate::vcl::Error> {
-        unsafe extern "C" fn chunk_collector<'a>(
+        unsafe extern "C" fn chunk_collector(
```
nyurik, line 130: Check if this change is correct

gquintard: if it compiles, I trust rust :-D

but yeah, it looks like it can be elided

gquintard:
looks like it broke the build

nyurik:
thx, yeppii - we have working CI :)

PR #37, opened 2024-09-04
Fix trait order and type

nyurik:
Make ordering of the trait impl the same as in definition, and modify return to prevent accidental mismatches.

### DIFF
--- a/examples/vmod_vdp/src/lib.rs
+++ b/examples/vmod_vdp/src/lib.rs
@@ -1,5 +1,7 @@
 varnish::boilerplate!();
 
+use std::ffi::CStr;
+
 use varnish::ffi;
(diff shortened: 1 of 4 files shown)


examples/vmod_vdp/src/lib.rs
```diff
@@ -41,6 +38,11 @@ impl VDP for Flipper {
         // send
         ctx.push(act, &self.body)
     }
+
+    // return our id, adding the NULL character to avoid confusing the C layer
```
gquintard, line 42:
```suggestion
    // return our id
```

gquintard:
it does change the API, but I'm in favor

nyurik:
heh, there are a lot of API changes as is ;)  But yes, for a crate that still uses 0.0.x version, API changes is no biggie :)

PR #38, opened 2024-09-04
rename `varnish_sys` into `ffi`

nyurik:
this way the naming is cleaner and non-redundant.

To rename: search replace `varnish_sys` to `ffi` and run `cargo +nightly fmt --all -- --config imports_granularity=Module,group_imports=StdExternalCrate` to reformat.


gquintard:
naming is hard, eh? ;-)

PR #40, opened 2024-09-05
spelling, minor styling

nyurik:
* please check what you meant by `Instand` in `examples/vmod_timestamp/src/lib.rs`
* `examples/vmod_example/README.md` has a non-existing ref to `src/vmod.rs`?
* lots of spell-fixes
* renamed `mutexed_hash_map` to `data` -- a bit cleaner, and it's private anyway


examples/vmod_timestamp/src/lib.rs
```diff
@@ -7,7 +7,7 @@ use varnish::vcl::vpriv::VPriv;
 
 varnish::vtc!(test01);
 
-// VPriv can wrap any (possibly custom) struct, here we only need an Instand from std::time.
+// VPriv can wrap any (possibly custom) struct, here we only need an instant from std::time.
```
gquintard, line 10:
```suggestion
// VPriv can wrap any (possibly custom) struct, here we only need an Instant from std::time.
```

examples/vmod_object/src/lib.rs
```diff
@@ -17,7 +17,7 @@ const EMPTY_STRING: String = String::new();
 
 // kv only contains one element: a mutex wrapping a String->String hashmap
 pub struct kv {
-    mutexed_hash_map: Mutex<HashMap<String, String>>,
+    data: Mutex<HashMap<String, String>>,
```
gquintard, line 20: since it's an example, I'd rather keep the name, just to make it painfully clear what that thing is to newcomers

gquintard:
two nitpicks, but all good otherwise

nyurik:
fixed

PR #41, opened 2024-09-05
more lints

nyurik:
* Simplified `HTTP::new`
* Please check `header_from_hd` -- it seems `unwrap()` is not needed?
* `std::ffi` should be used instead of `std::os::raw` per [docs](https://doc.rust-lang.org/std/os/raw/)
* a few minor fixes like use statements, asserts, and doc links

### DIFF
--- a/src/vcl/http.rs
+++ b/src/vcl/http.rs
@@ -38,13 +38,9 @@ pub struct HTTP<'a> {
 impl<'a> HTTP<'a> {
     /// Wrap a raw pointer into an object we can use.
     pub fn new(p: *mut ffi::http) -> Option<Self> {
-        if p.is_null() {
-            None
-        } else {
-            Some(HTTP {
-                raw: unsafe { p.as_mut().unwrap() },
-            })
-        }
+        Some(HTTP {
+            raw: unsafe { p.as_mut()? },
+        })
     }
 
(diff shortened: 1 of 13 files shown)


src/vcl/http.rs
```diff
@@ -263,7 +259,7 @@ fn header_from_hd<'a>(txt: *const ffi::txt) -> Option<(&'a str, &'a str)> {
         let e = (*txt).e;
         let buf = from_raw_parts(b.cast::<u8>(), e.offset_from(b) as usize);
 
-        let colon = buf.iter().position(|x| *x == b':').unwrap();
+        let colon = buf.iter().position(|x| *x == b':')?;
```
gquintard, line 262: that one is an actual invariant, the lower levels guarantee it

nyurik: sure, but do we actually want to add exception throwing code here? I think it is easier to just return None if for some magical reason it has it?

gquintard: no, if that invariant isn't maintained, something is very wrong, possibly because some memory corruption, we should blow up early and loudly so we (`varnishd` really) can fix the issue. Hiding the problem will hurt us in the long run

nyurik: thanks, makes sense, fixed

PR #43, opened 2024-09-05
Set `VARNISHTEST_DURATION=5s` by default

nyurik:
This ensures varnishtest will fail much quicker by default.


gquintard:
please get us a changelog entry for that one, great addition, people need to be aware of it

nyurik:
thx, fixed

PR #45, opened 2024-09-05
Some CStr cleanup

nyurik:
* remove NULL-char related comments when using `CStr` - that's the definition of it.
* casting `cstr.as_ptr()` with `.cast::<c_char>()` is redundant - it's already that.
* modify `WS::copy_bytes_with_null` to return `CStr`, and marked it as `unsafe`. Current implementation relies on the input value to not contain any `\0` chars, and would result in UB otherwise.

See also #46

### DIFF
--- a/src/vcl/ws.rs
+++ b/src/vcl/ws.rs
@@ -10,7 +10,7 @@
 //! **Note:** unless you know what you are doing, you should probably just use the automatic type
 //! conversion provided by [`crate::vcl::convert`], or store things in
 //! [`crate::vcl::vpriv::VPriv`].
-use std::ffi::{c_char, c_void};
+use std::ffi::{c_char, c_void, CStr};
 use std::marker::PhantomData;
 use std::ptr;
 use std::slice::from_raw_parts_mut;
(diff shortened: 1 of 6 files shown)


src/vcl/ws.rs
```diff
@@ -86,14 +87,20 @@ impl<'a> WS<'a> {
...
+        src: &T,
+    ) -> Result<&'a CStr, String> {
         let buf = src.as_ref();
         let l = buf.len();
 
         let dest = self.alloc(l + 1)?;
         dest[..l].copy_from_slice(buf);
         dest[l] = b'\0';
-        Ok(dest)
+        Ok(unsafe { CStr::from_bytes_with_nul_unchecked(dest) })
```
gquintard, line 103: shouldn't we check for NULL and error out?

nyurik: @gquintard we should, but i would rather do it as part of the #46 fix.  For now this is more of a refactoring.  Baby steps :)

gquintard: yes, but this refactor both changes the API and creates another possibility of panicking.

why not just try to convert and bubble up the error?

nyurik: I added a null check right above - is that ok?

gquintard: oh, sorry, I missed that, I was expecting you to use `from_bytes_with_nul`, my bad

nyurik: i didn't want to use it after copying possibly bad data to the workspace

gquintard: fair point

PR #53, opened 2024-09-06
Simplify `Ok(v) => v`

nyurik:
`let else` and `unwrap_or` are shorter and cleaner replacements for handling error conditions.


src/lib.rs
```diff
@@ -230,17 +230,19 @@ macro_rules! vtc {
 macro_rules! boilerplate {
     () => {
         #[allow(
-            non_upper_case_globals,
+            dead_code,
             non_camel_case_types,
             non_snake_case,
+            non_upper_case_globals,
```
gquintard, line 236: shouldn't we be removing those?

nyurik: I simply sorted it - we should remove some/most of these, just not yet

gquintard: but `unused_mut` or is it the pre-coffee haze hindering me? Anyway, ok with me

nyurik: I moved `unused_mut` from inside the python-generated script - no point in having some allows generated for every method, while others are here.  Once proc-macro lands, it will become much simpler to fix these

PR #54, opened 2024-09-06
Merge 1st: Proc macro implementation

### LINKED ISSUE #73: New VPriv support
I came up with this model - showing the changes needed in the `varnish` crate, changes to the generated code, and how it will be used by the user.  Note that for the user it will probably need some attributes too.

Most important change:  this completelly gets rid of `VPriv` and `VPrivInt` structs.

```rust
//
// This is part of the Varnish core Rust lib
//
mod varnish {
    use std::ffi::c_void;
    use std::ptr;
    use std::ptr::null;

    use crate::ffi;

    /// SAFETY: ensured by Varnish itself
    unsafe impl Sync for ffi::vmod_priv_methods {}

    /// Take ownership of the object of type `T` and return it as a `Box<T>`.
    /// The original pointer is set to null.
    ///
    /// SAFETY: `priv_` must reference a valid `T` object pointer or `NULL`
    unsafe fn get_owned_bbox<T>(priv_: &mut *mut c_void) -> Option<Box<T>> {
        let obj = ptr::replace(priv_, ptr::null_mut());
        if obj.is_null() {
            None
        } else {
            Some(Box::from_raw(obj.cast::<T>()))
        }
    }

    /// A Varnish callback function to free a `vmod_priv` object.
    /// Here we take the ownership and immediately drop the object of type `T`.
    /// Note that here we get `*priv_` directly, not the `*vmod_priv`
    ///
    /// SAFETY: `priv_` must be a valid pointer to a `T` object or `NULL`.
    pub unsafe extern "C" fn free_vmod_priv<T>(_ctx: *const ffi::vrt_ctx, mut priv_: *mut c_void) {
        // we get the ownership and don't do anything with it, so it gets dropped at the end of this fn
        get_owned_bbox::<T>(&mut priv_);
    }

    impl ffi::vmod_priv {
        /// Transfer ownership of the object to the caller, cleaning up the internal state.
        ///
        /// SAFETY: `priv_` must reference a valid `T` object pointer or `NULL`
        pub unsafe fn take<T>(&mut self) -> Option<Box<T>> {
            self.methods = null();
            get_owned_bbox(&mut self.priv_)
        }

        /// Set the object and methods for the `vmod_priv`, and the corresponding static methods.
        ///
        /// SAFETY: The type of `obj` must match the type of the function pointers in `methods`.
        pub unsafe fn put<T>(&mut self, obj: Box<T>, methods: &'static ffi::vmod_priv_methods) {
            self.priv_ = Box::into_raw(obj).cast();
            self.methods = methods;
        }

        /// Use the object as a reference, without taking ownership.
        ///
        /// SAFETY:
        /// * `priv_` must reference a valid `T` object pointer or `NULL`
        /// * Assumes the pointer stored in `priv_` can be re-interpreted as a &Option<Box<T>>
        /// * `take()` must not be called on the same `vmod_priv` object until the returned reference is dropped
        /// * cleanup must not be done on the object until the returned reference is dropped
        pub unsafe fn get_ref<T>(&self) -> Option<&T> {
            Some(self.priv_.cast::<Box<T>>().as_ref()?)
        }
    }
}

//
// This is generated with proc macro based on the user code (see below)
//
pub mod generated {
    use super::user_code;
    use crate::ffi;
    use crate::vpriv_demo::varnish::free_vmod_priv;

    /// Static methods to clean up the `vmod_priv` object's `user_code::UserType`
    static MY_TYPE_METHODS: ffi::vmod_priv_methods = ffi::vmod_priv_methods {
        magic: ffi::VMOD_PRIV_METHODS_MAGIC,
        type_: c"user_code::UserType".as_ptr(),
        fini: Some(free_vmod_priv::<user_code::UserType>),
    };

    /// Called by Varnish as a wrapper around the user's mutable function
    pub unsafe fn user_mut_fn_wrapper(vp: *mut ffi::vmod_priv) {
        // Take ownership of the user's object and call the user's function with it
        let mut obj = (*vp).take();
        user_code::user_mut_func(&mut obj);
        if let Some(obj) = obj {
            // Release ownership back to Varnish
            (*vp).put(obj, &MY_TYPE_METHODS);
        }
    }

    /// Called by Varnish in a read-only context
    pub unsafe fn user_fn_wrapper(vp: *const ffi::vmod_priv) -> i32 {
        // Do not take ownership - multiple simultaneous calls are allowed
        user_code::user_func((*vp).get_ref())
    }
}

//
// This is what a user could write
//
pub mod user_code {
    /// Example user type
    pub struct UserType {
        pub data: i32,
    }

    /// Mutable access to the data field of UserType, instantiating it if needed
    /// Rust ensures this function is the only one that owns UserType
    pub fn user_mut_func(obj: &mut Option<Box<UserType>>) {
        if let Some(obj) = obj {
            obj.data += 1;
        } else {
            *obj = Some(Box::new(UserType { data: 1 }));
        }
    }

    /// Read-only access to the data field of UserType.
    /// Multiple refs to the same Box could exist, and can contain Mutex or RwLock if needed.
    pub fn user_func(obj: Option<&UserType>) -> i32 {
        obj.as_ref().map_or(0, |obj| obj.data)
    }
}
```

nyurik:
Implement procedure macro support, replacing the need for the build script and the boilerplate code.  Includes a testing framework (using [insta](https://insta.rs/)), reports multiple compiler errors linking to offending code with tests using [trybuild](https://docs.rs/trybuild/latest/trybuild/), and autogenerated documentation.

All examples have been migrated in the #85 

## Usage example

```rust
/// Attaching `vmod` attribute to a `mod` exports its content as VMOD functions, methods, and objects
/// The name of VMOD is the name of this module.
/// If `docs = "file"` param is set, a documentation file will be auto-generated on each build.
/// The path is relative to the root of the project.
#[varnish::vmod(docs = "README.md"]
mod my_vmod {
    /// Simple function exported to VCL with a single mandatory param
    pub fn simple(val: i64) -> bool {
        todo!()
    }

    /// Simple function exported to VCL with a positional and an optional param
    pub fn optional(val: i64, opt: Option<i64>) -> bool {
        todo!()
    }

    /// A function that requests a few additional values automatically provided.
    /// Note that this function is identical to the previous one from the VCL perspective.
    pub fn complex(
        /// The context of the current VCL call
        ctx: &Ctx,
        /// Readonly shared struct. Only one "per_vcl" type is allowed per VMOD
        #[shared_per_vcl] shared1: Option<&MyPerVclStruct>,
        /// Mutable per-task struct. Only one "per_task" type is allowed per VMOD
        #[shared_per_task] shared2: &mut Option<Box<MyPerTaskStruct>>,
        /// Same params as the previous function 
        val: i64,
        opt: Option<i64>,
    ) -> Result<i64, &str> {
        todo!()
    }

    /// On VCL event like loading, call this function in an exclusive context. Must be tagged with `#[event]`
    #[event] pub fn on_event(
        /// Same type as in complex, but this time we get exclusive access to it
        #[shared_per_vcl] val: &mut Option<Box<sub::MyPerVclStruct>>,
        /// What event caused this function to be called
        evt: Event,
    ) -> i64 { todo!() }

    /// All these functions will work as methods for the object `MyObject`
    /// The actual object can be declared in another module to reduce clutter in the exported mod.
    impl MyObject {
        /// Constructor for the object. Must be tagged with `#[constructor]`
        pub fn new(
            /// The name of the object in VCL - this is auto-provided, not part of params in VCL
            #[vcl_name] name: &str,
            /// This value is passed from VCL: `new foo = my_vmod.MyObject(42);
            val: i64,
        ) -> Self {
            todo!()
        }

        /// Method on the object - same as regular functions, but with an additional `self` param
        pub fn method(&self, val: i64) -> i64 {
            todo!()
        }
    }
}

/// The actual struct and other non-public functionality should be outside of `mod my_vmod`
struct MyObject;

impl MyObject {
    pub fn non_exported_functions() { todo!() }
}
```

## Converted Examples
* [x] vmod_test
* [x] vmod_be
* [x] vmod_error
* [x] vmod_event
* [x] vmod_example
* [x] vmod_infiniteloop
* [x] vmod_object
* [x] vmod_timestamp
* [x] vmod_vdp
* [x] vmod_vfp

## Notes
* Attribute must be on a module because we need to generate all functions together, and proc macros do not support state sharing between separate macro calls. See https://github.com/rust-lang/rust/issues/44034
* There is no need for boilerplate macros or the build step
* `Ctx` is now optional, and can be both `&` and `&mut`
* This PR includes refactoring of the `Cargo.toml` done in #62

## Tasks
* [x] Generate `extern "C"` wrapper for each fn (partially done, needs a bit more cleanup)
* [x] Generate struct for optional parameter passing Varnish->Wrapper
* [x] Generate `struct VmodExports` with all functions as static callbacks, and instantiate it as global static `VMOD_EXPORTS`.
* [x] Generate `$CPROTO` C header snippet describing function signatures
* [x] Generate description JSON
* [x] Generate `Vmod_{name}_Data` global static instance
* [x] Object support
* [x] Optional types
* [x] Generate documentation
* [x] Report multiple errors at once, with ui tests
* [x] Support default values, e.g. `fn foo(#[arg(default = "arg")], param: &str)`

### Other Ideas and TODOs
* [ ] Support for copying result into workspace and returning a ref to it, instead of double-copying
* [ ] Nullable types like `SocketAddr`, `Probe`, and `COWProbe` create a bit of confusion because they do not have to be "optional", they can just be nullable, and in Rust nullable and optional is the same thing. This creates a problem if the user wants a function like `foo(ip: SocketAddr, val: i64)` -- the first optional won't work properly unless the second is also optional.

### Supported types

* Context type `&Ctx` or `&mut Ctx` as input param is supported in all user functions
* VCL name can be taken as a `&str` input param for object constructors. Requires `#[vcl_name]` attribute.

|     Input      | Output |             Rust              |     Varnish     |             C type              |
|:--------------:|:------:|:-----------------------------:|:---------------:|:-------------------------------:|
|                |        |                               |    `VCL_ACL`    |        `*const vrt_acl`         |
|                |   ✅    |        `VCLBackendPtr`        |  `VCL_BACKEND`  |        `*const director`        |
|                |        |                               |   `VCL_BLOB`    |        `*const vrt_blob`        |
|                |        |                               |   `VCL_BODY`    | `*const ::std::os::raw::c_void` |
|       ✅        |   ✅    |            `bool`             |   `VCL_BOOL`    |    `::std::os::raw::c_uint`     |
|                |        |                               |   `VCL_BYTES`   |                                 |
|       ✅        |   ✅    |          `Duration`           | `VCL_DURATION`  |           `vtim_dur`            |
|                |        |                               |   `VCL_ENUM`    | `*const ::std::os::raw::c_char` |
|                |        |                               |  `VCL_HEADER`   |        `*const gethdr_s`        |
|                |        |                               |   `VCL_HTTP`    |           `*mut http`           |
|                |        |                               | `VCL_INSTANCE`  |    `::std::os::raw::c_void`     |
|       ✅        |   ✅    |             `i64`             |    `VCL_INT`    |                                 |
| ✅ <sup>2</sup> |   ✅    |         `SocketAddr`          |    `VCL_IP`     |        `*const suckaddr`        |
| ✅ <sup>2</sup> |   ✅    |            `Probe`            |   `VCL_PROBE`   |   `*const vrt_backend_probe`    |
| ✅ <sup>2</sup> |   ✅    |          `COWProbe`           |        "        |                "                |
|       ✅        |   ✅    |             `f64`             |   `VCL_REAL`    |                                 |
|                |        |                               |   `VCL_REGEX`   |          `*const vre`           |
|                |        |                               | `VCL_STEVEDORE` |       `*const stevedore`        |
|                |        |                               |  `VCL_STRANDS`  |        `*const strands`         |
|                |   ✅    |         `VCL_STRING`          |  `VCL_STRING`   | `*const ::std::os::raw::c_char` |
| ✅ <sup>1</sup> |   ✅    |            `&str`             |        "        |                "                |
|                |   ✅    |           `String`            |        "        |                "                |
|                |        |            `&CStr`            |        "        |                "                |
|                |        |           `CString`           |        "        |                "                |
|                |        |                               |    `VCL_SUB`    |        `*const vcl_sub`         |
|                |        |                               |   `VCL_TIME`    |           `vtim_real`           |
|                |        |                               |    `VCL_VCL`    |           `*mut vcl`            |
|      n/a       |   ✅    |                               |   `VCL_VOID`    |    `::std::os::raw::c_void`     |
|                |        |             `i8`              |                 |                                 |
|                |        |             `i16`             |                 |                                 |
|                |        |             `i32`             |                 |                                 |
|                |        |            `i128`             |                 |                                 |
|                |        |            `isize`            |                 |                                 |
|                |        |             `u8`              |                 |                                 |
|                |        |             `u16`             |                 |                                 |
|                |        |             `u32`             |                 |                                 |
|                |        |             `u64`             |                 |                                 |
|                |        |            `u128`             |                 |                                 |
|                |        |            `usize`            |                 |                                 |
|                |        |            `[u8]`             |                 |                                 |
|                |        |           `Vec<u8>`           |                 |                                 |
|                |   ✅    | `Box<dyn Error>` <sup>3</sup> |                 |                                 |

* <sup>1</sup> - these types support `#[arg(default = ...)]` attribute, allowing a function to be called from VCL with this argument set to the VMOD-provided default.
* <sup>2</sup> - these types must be declared as `Option<T>` in Rust code, making it both optional and nullable, e.g. `Option<SocketAddr>`.
* <sup>3</sup> - `Box<dyn Error>` is a special case, as it is not a Varnish type, but it is a common return type for error handling in Rust. It is supported as an error return type, but not as an input or output type.

Closes #73


nyurik:
no idea why this is failing tbh, any thoughts?

gquintard:
problem with finding the JSON description apparently:
```
**** v1    CLI RX|\tFile name: /home/runner/work/varnish-rs/varnish-rs/target/debug/deps/libvmod_example.so
**** v1    CLI RX|\tError: No VMOD JSON found
**** v1    CLI RX|('<vcl.inline>' Line 5 Pos 16)
**** v1    CLI RX|        import example from "/home/runner/work/varnish-rs/varnish-rs/target/debug/deps/libvmod_example.so";
**** v1    CLI RX|---------------#######-------------------------------------------------------------------------------------
**** v1    CLI RX|
**** v1    CLI RX|Running VCC-compiler failed, exited with 2
```

examples/vmod_be/src/lib.rs
```diff
@@ -1,42 +1,46 @@
...
-        self.be.vcl_ptr()
+    impl parrot {
+        pub fn new(ctx: &mut Ctx, to_repeat: &str) -> Result<Self, String> {
+            // to create the backend, we need:
+            // - the vcl context, that we just pass along
+            // - the vcl_name (how the vcl writer named the object)
+            // - a struct that implements the Serve trait
+            let be = Backend::new(
+                ctx,
+                "TODO", // FIXME: This should be the vcl_name param that currently is not forwarded
```
gquintard, line 30: that is important because that allows us to name the backend/processors we create

examples/vmod_be/src/lib.rs
```diff
@@ -1,42 +1,46 @@
...
+    // - backend(), so that we can produce a C pointer for varnish to use
+    #[allow(non_camel_case_types)]
+    pub struct parrot {
+        be: Backend<Sentence, Body>,
     }
 
-    pub fn backend(&self, _ctx: &Ctx) -> VCLBackendPtr {
-        self.be.vcl_ptr()
+    impl parrot {
+        pub fn new(ctx: &mut Ctx, to_repeat: &str) -> Result<Self, String> {
```
gquintard, line 23: why the error change? the previous signature was very convenient

examples/vmod_event/src/lib.rs
```diff
@@ -1,38 +1,62 @@
...
-static mut N: i64 = 0;
+    pub fn loaded(#[shared_per_vcl] shared: Option<&i64>) -> i64 {
+        shared.copied().unwrap_or(0)
+    }
 
-pub fn loaded(_: &Ctx, vp: &VPriv<i64>) -> i64 {
-    // unwrapping is safe here as we know that the event function will have loaded something
-    *vp.as_ref().unwrap()
-}
+    /// FIXME: does Varnish guarantee that `loaded()` will never be called while `on_event()` is running?
```
gquintard, line 17: yes

examples/vmod_event/src/lib.rs
```diff
@@ -1,38 +1,62 @@
...
+    ) -> Result<i64, &'static str> {
+        // log the event, showing that it implements Debug
+        ctx.log(LogTag::Debug, format!("event: {event:?}"));
+
+        // we only care about load events, which is why we don't use `match`
+        if matches!(event, Event::Load) {
+            // increment the count in a thread-safe way
+            let last_count = EVENT_LOADED_COUNT.fetch_add(1, Relaxed);
+            if last_count == 1 {
+                // FIXME: why, what reasons?
```
gquintard, line 34: for demo purposes

examples/vmod_event/src/lib.rs
```diff
@@ -1,38 +1,62 @@
...
+                // fail the second load, because reasons
+                // FIXME: what should VCL wrapper return in case of an error?
```
gquintard, line 36: don't you like my string?

examples/vmod_event/src/lib.rs
```diff
@@ -1,38 +1,62 @@
...
+        ctx: &mut Ctx,
+        #[shared_per_vcl] shared: &mut Option<Box<i64>>,
+        event: Event,
+    ) -> Result<i64, &'static str> {
+        // log the event, showing that it implements Debug
+        ctx.log(LogTag::Debug, format!("event: {event:?}"));
+
+        // we only care about load events, which is why we don't use `match`
+        if matches!(event, Event::Load) {
+            // increment the count in a thread-safe way
```
gquintard, line 31: it doesn't have to be thread-safe, there's only ever one thread reading it. Because the compiler can't prove it doesn't mean it's unsafe. I'm completely fine with escaping with `unsafe` here.

examples/vmod_example/src/lib.rs
```diff
@@ -1,60 +1,53 @@
...
-        Some(n) => format!("I was given {n} as argument"),
+    /// This function can be called without arguments, or with an integer:
+    ///
+    /// ```vcl
+    /// set resp.http.Obvious = example.captain_obvious();
+    /// set resp.http.Obvious-Number = example.captain_obvious(42);
+    /// ```
+    ///
+    /// CString is a Rust's UTF-8 encoded null-terminated string,
+    /// making it the most efficient way to pass strings to Varnish.
```
gquintard, line 18: that's not exactly true, the best way is to create a `&CStr` on the workspace and return that

examples/vmod_vfp/src/lib.rs
```diff
@@ -1,13 +1,45 @@
...
+use varnish::vcl::processor::{InitResult, PullResult, VFPCtx, VFP};
+
+#[varnish::vmod]
+mod vfp {
+    use varnish::ffi;
+    use varnish::vcl::ctx::{Ctx, Event};
+    use varnish::vcl::processor::new_vfp;
 
-varnish::vtc!(test01);
+    use crate::Lower;
```
gquintard, line 12: could we have kept `Lower` at the same level as the rest of the code?

gquintard (changes requested):
I have a bunch of comments, but the two most important ones are:
- I'd really like to be able to generate a some md/rst docs for the API, I don't want people to have to read the code
- we really want the `vcl-name` to be passed to the object constructor

I need to play with the `varnish-macro` crate, it's a hug piece of code and I don't understand it yet, this needs to change. Notably I want to reassure myself that we are properly cleaning `vmod_priv`s, and what this approach means for the existing vmods.

Also, please transfer any API descriptions that may appear in the `vcc` files, and then `git rm` the files. If we are moving away from them, let's do it fully.

examples/vmod_event/src/lib.rs
nyurik: First off, most likely this has zero cost on performance - most such constructs are compiled away.

That said, I would recommend against `unsafe` for several reasons:
* it teaches good patterns to possibly-novice users, making them think it is not a big deal to use `unsafe`
* sprinkling `unsafe` effectively makes the entire code unsafe -- I read somewhere a good example of how one unsafe and unsound function would make the rest of the safe rust API be unsound
* user may copy/paste that code elsewhere where they might not understand the implications - resulting in unsound behavior

The whole idea of Rust is that we can code without "thinking" - the compiler does safety checks for us.  Adding unsafe means you have to constantly pay attention to it, which is not common for most Rust devs (that's the who idea of wrapping unsafe API by a small team so that bigger team can treat it as safe).  In short - it would have been safer to write this code in C because at least people are trained to think in terms of safety there.


For now, we still have to use `unsafe` in a few places - lets limit them and try to wrap them cleanly too rather than teaching this.

examples/vmod_event/src/lib.rs
nyurik: lol, i updated it a bit now that i understand its for demoing (for a moment i thought it was some internal idiosyncrasy

examples/vmod_example/src/lib.rs
nyurik: I wrote it when trying to get CStrings to work as return types, they kinda do, but not worth it now that we use `txt` internally. Removed.

examples/vmod_vfp/src/lib.rs
nyurik: The whole module must be tagged with `#[vmod]` attribute in order for the proc-macro to modify it.  The macro currently looks for several signals:
* a public function -- treated as a function or an event if it has `#[event]`
* an `impl` block -- treated as methods (only allows non-trait impls)

This means that it is usually cleaner to separate Varnish interface from the rest of the code, and this pattern demos how to do that. We could have more code inside the `vfp` mod, but it might create some surprises or simply won't compile in some cases, e.g. some public fn without the right signature - so better teach best practices this way

examples/vmod_be/src/lib.rs
nyurik: Implemented

gquintard: not that it's not really magic or anything, in the C world, `ctx` and `vcl_name` are the first arguments of the constructor. Maybe it would be easier to just force them to be in the `rust` world too?

examples/vmod_be/src/lib.rs
nyurik: `Box<dyn Error>` is now supported

nyurik:
All issues including `vcl-name` and `Box<dyn Error>` have been fixed, with the only open question remaining is the documentation.  I moved all docs from `*.vcc` to the rust doc comments (the `///` comments on `mod` and `fn` elements).

As far as I can tell, the current code does not automatically generate documentation as part of `cargo build --workspace --all-targets`.  Even though `vmodtool-rs.py` has a `-w` / `---rstdir` param, and it gets set to the `OUT_DIR` by each example's build script, there are no doc files generated. Could this be a silent failure due to a missing python lib somewhere?  I couldn't find any docs on that.

Now, as far as doc gen, we need to come up with a good strategy.

* The proc macro has access to the `///` comments while parsing, and can modify them (they are actually [`#doc`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html) attributes in the parse tree). So we can auto-generate some info on what the function will look like from VCL:
  * `fn my_func(ctx: &Ctx, #[shared_per_vcl] shared: ..., foo: String, bar: Option<i64>) -> ...` can be documented as "Use this function as `my_func(foo, [bar])`, where `foo` is a string, and `bar` is an optional integer."
  * `fn new(ctx: &Ctx, ...) -> Self` as "Creates an instance of the object `objname`. In VCL, use `new objname(...)`

* We can rely on tools like [cargo-readme](https://github.com/webern/cargo-readme) or others to actually generate the needed docs, with the only problem being how to filter the relevant data out of the crate.

nyurik:
documentation is now autogenerated. All examples use `#[vmod(docs = "VMOD.md")]` to auto-regenerate that file on each build.  I am a bit unsure if it should generate README or not, because it will auto-regenerate it, and if the user changes it before committing, it will be lost.  One option is to add a hidden comment to the .md file at the top, but its not a foolproof way either...

gquintard:
as the files are generated, I would say we don't need to commit them, we can even put them in `.gitignore`.

They do look awesome though!

gquintard:
I'm completely ok with a one-liner at the top saying it's autogenerated, as you said, it's not foolproof, but it's good enough

nyurik:
ok, updated, and renamed. This will force users to maintain all their documentation inside the code :)

gquintard:
it's for their own good™

varnish-macros/src/generator.rs
```diff
@@ -178,6 +178,7 @@ impl Generator {
                     VCL_DURATION,
                     VCL_INT,
                     VCL_IP,
+                    VCL_PROBE,
```
gquintard, line 181: Duuuuuuuuee, just landed and implemented that one while flying as `reqwest` needs it.

Welp, the good thing is that I'm pretty close to what you did here \o/

nyurik: LOL, the famous "merge conflict after flying" issues :)

I just hacked it in a few minutes trying to figure out if it will work or not.  I don't think its much of a  waste because you now have far better understanding of how it works :)  The issue though is that `Option` in Rust could mean both `nullable` and `optional` from Varnish perspective - and unless Varnish can ensure values like `SocketAddr` are always non-null, we will have to do some heuristics plus some extra attributes... Not sure though.

nyurik:
These types are the only references allowed as inputs at the moment: `str`, `Probe`, `COWProbe`, `SocketAddr`.  All of them are essentially "nullable" - allowing user's VCL code to pass in a `NULL`.  `&str` is the only one allowed to be declared without the `Option<&str>` wrapper, which creates a bit of a confusion:
* If VCL allows ALL values to be NULLs, shouldn't `&str` require to be `Option<&str>`?  Crashing is never a good thing on a missing/null argument, so if it's a valid input from VCL, it should be declared as such, and let user's code deal with it.
* The positional (required) arguments is different from the nullable ones.  `Option<&T>` in our case means both. As discussed, we will add the `#[arg(required)]` attribute flag to indicate the rarer case of **nullable** but not **optional**.  It is an error to add this flag to the types without the `Option` wrapper, or to the types that are not nullable. These are invalid to have the `required` flag:  `Option<i64>` or `i64`

nyurik:

PR #60, opened 2024-09-09
Make `ffi::Txt` safer and simpler to use

nyurik:
Add a few helper methods to help with converting txt to str and back, which also helps with the header parsing.  As a tiny addition, also simplified `TestCtx` instantiation.

### DIFF
--- a/src/vcl/utils.rs
+++ b/src/vcl/utils.rs
@@ -1,8 +1,66 @@
+use std::ffi::{c_char, CStr};
+use std::slice::from_raw_parts;
+use std::str::from_utf8;
+
 use crate::ffi::{
-    director, req, sess, vcldir, vfp_ctx, vfp_entry, vrt_ctx, ws, DIRECTOR_MAGIC, REQ_MAGIC,
+    director, req, sess, txt, vcldir, vfp_ctx, vfp_entry, vrt_ctx, ws, DIRECTOR_MAGIC, REQ_MAGIC,
     SESS_MAGIC, VCLDIR_MAGIC, VFP_CTX_MAGIC, VFP_ENTRY_MAGIC, VRT_CTX_MAGIC, WS_MAGIC,
 };
 use crate::vcl::backend::{Serve, Transfer, VCLBackendPtr};
+
+impl txt {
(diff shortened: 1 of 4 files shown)


src/vcl/utils.rs
```diff
@@ -1,7 +1,66 @@
...
+        Self::from_bytes(s.as_bytes())
+    }
+
+    pub fn from_cstr(s: &CStr) -> Self {
+        Self::from_bytes(s.to_bytes())
+    }
+
+    /// Convert the `txt` struct to a `&[u8]`.
+    /// FIXME: Consider modifying this to return an empty slice instead of None,
+    ///        Unless we want to explicitly differentiate between empty and null.
```
gquintard, line 31: we absolutely want to distinguish between empty and null
```suggestion
```

src/vcl/utils.rs
```diff
@@ -1,7 +1,66 @@
...
+        self.to_slice().map(|s| from_utf8(s).unwrap())
+    }
+
+    /// Parse the `txt` struct as a header, returning a tuple with the key and value,
+    /// trimming the value of leading whitespace.
+    pub fn parse_header<'a>(&self) -> Option<(&'a str, &'a str)> {
+        // We expect varnishd to always given us a string with a ':' in it
+        // If it's not the case, blow up as it might be a sign of a bigger problem.
+        let (key, value) = self.to_str()?.split_once(':').unwrap();
+        // FIXME: Consider `.trim_ascii_start()` if unicode is not a concern
```
gquintard, line 59: header names will be ascii, so go for it

nyurik: header name is not the issue - `.trim_start` uses `char::is_whitespace`, whereas `.trim_ascii_start` only removes "ascii whitespace" by using `u8::is_ascii_whitespace`.  Which means that if a header **value** can have non-ascii whitespace characters in the front, they will stay there.  Let's merge and decide this separate?

src/vcl/utils.rs
nyurik: thx, I updated the comment

PR #61, opened 2024-09-09
Minor `VPriv` improvements

nyurik:
* Make `Vpriv` store a reference instead of a pointer, with an attached lifetime.  Unfortunately we have to support `into_rust` which must fake it with `'static`.
* Do not double-wrap name string - first as a `CString` inside a `Box` because `CString` is already heap-allocated.
* Improve `store` fn

### DIFF
--- a/src/vcl/convert.rs
+++ b/src/vcl/convert.rs
@@ -435,9 +435,11 @@ impl IntoRust<Duration> for VCL_DURATION {
     }
 }
 
-impl<T> IntoRust<VPriv<T>> for *mut vmod_priv {
-    fn into_rust(self) -> VPriv<T> {
-        VPriv::<T>::new(self)
+impl<'a, T> IntoRust<VPriv<'a, T>> for *mut vmod_priv {
+    fn into_rust(self) -> VPriv<'a, T> {
+        // FIXME: this is not a good pattern for sure,
+        // but we assume that vmod_priv will live longer than VPriv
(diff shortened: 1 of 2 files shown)


nyurik:
@gquintard I think I am approaching this one wrong.  Looking at your comment

```
// This is annoying. `vmod_priv` contains a pointer to `vmod_priv_methods`
// that we need to use to free our object, however, we need to piggy-back
// on vmod_priv_methods->free to also clean vmod_priv_methods. So we create
// an InnerVPriv that encapsulate both our object, and vmod_priv_methods.
// Ideally:
// - vmod_priv_methods would have a method to free itself
// - we would be able to create vmod_priv_methods from a const fn
```

It seems you are suggesting something like this (?):

```rust
impl Drop for ffi::vmod_priv_methods {
  // magic
}
```

Moreover, you re-use the same `name` object, referencing it from both `ffi::vmod_priv_methods` and `InnerVPriv` - somehow this seems like a potential for a bug?  Not certain, may need more explanation of what is going on

gquintard:
the C layer will call us, checking if `vmod_priv->free` exists, and if it does, calls it, so we have that incestuous relationship with the object pointing at its methods who also must clean it.

Thus I don't think the `Drop` implementation is buying us anything here.

as for the name, I think we can actually remove it from the `InnerPriv`, I don't see much of a reason to have it here

PR #63, opened 2024-09-12
Add debug mode to the `vtc!` macro

nyurik:
`vtc!` macro now supported debug mode, e.g. `vtc!(file, true)` - keeps the temporary files and always prints the output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unpublished
 
 - `vtc!` macro now supports `VARNISHTEST_DURATION` env var, defaulting to "5s"
+- `vtc!` macro now supported debug mode, e.g. `vtc!(file, true)` - keeps the temporary files and always prints the output.
 
 # 0.0.19 (2024-03-24)
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -171,9 +171,17 @@ pub mod vcl {
 /// otherwise you'll get a panic.
 ///
 /// Tests will automatically time out after 5s. To override, set `VARNISHTEST_DURATION` env var.
+///
+/// To debug the tests, pass `true` as the second argument:
+/// ``` rust
+/// varnish::vtc!(test01, true);
+/// ```
 #[macro_export]
 macro_rules! vtc {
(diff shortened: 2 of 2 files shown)


gquintard:
I can imagine that nacro growing quite a bit, considering all the options `varnishtest` has. Would it make sense to use `keep=true` or something like that?

nyurik:
I am not certain how to do this better. How about this:
* A simple new macro `vtc_test_debug!(file)` which uses the one below with extra params
* Add more params: `vtc_test!(file, print_on_pass = true, args = "-L -v")` - allows any options. Unfortunately, macro params are finicky, so params must be in that order and cannot be skipped, or macro declaration will be somewhat long

nyurik:
P.S. I feel there is a bit too much magic in this macro (this is an unrelated discussion to the above, but since we are on the topic...):
* `vtc!` is non-descriptive - perhaps rename it to `vtc_test!`?
* The macro is too magical -- it should not auto-magically modify paths or add extensions -- Rust has a lot of macros that work with files, e.g. `include_str!`, and they all require a string, where that string is relative to the current file.  So `vtc_test!("../tests/test01.vtc"` seems more readable, but a bit more verbose.
* All that code to execute `varnishtest` should be a proper function, not a macro expansion.  Macro should call it.

gquintard:
all good points, and I'm open to them all.

How about this: we keep in mind that `vtc!()` will need to be deprecated, and in the meantime, we add a single argument to it: a string that is just added to the `varnishtest` arguments.

So, if you do want to keep the data around, `vtc!(foo.vtc, "-L")`. This way, you get to keep your debug files, but we don't have to add a thousand different switches before we phase out `vtc!`

nyurik:
I am not sure adding a string is a good path because you won't know which `/tmp/...` dir is being used in case of a success -- the output is not shown unless it crashes. That's why i wanted to do a simple "debug=true` or `vtc_debug!(...)` style command - so that users don't have to learn all the vcttest CLI arguments, and they don't forget to turn it off because they always see the output if it is saved.  Esp if its going away?

gquintard:
fair argument, merging

PR #68, opened 2024-09-13
Efficient logging without extra allocations

nyurik:
All logging functions now take a new `Loggable` object that contains a `Cow<CStr>` - making it possible to log without an extra re-allocation.


gquintard:
I've been pondering this new type? We should validate that there's no `\0` in the buffer, but we can print `String` and `&str` using `%.*s`, this wouldn't change the API and still doesn't need extra allocations

nyurik:
not sure what you mean - are you saying you are ok to send garbage to the print function?  Note that currently we do assert the string is valid, e.g. `let c_cstring = CString::new(msg).unwrap();` in `fail()`

PR #69, opened 2024-09-13
Optimize log usage - take 2

nyurik:
This replaces #68 without the extra struct


src/vcl/ctx.rs
```diff
@@ -220,15 +220,15 @@ impl Event {
     }
 }
 
-// TODO: avoid Stringification
-pub fn log(logtag: LogTag, msg: &str) {
+pub fn log(logtag: LogTag, msg: impl AsRef<str>) {
+    let msg = msg.as_ref();
```
gquintard, line 224: Id prefer to use `VSLbt` here, but I can change it post-merge

PR #71, opened 2024-09-14
Minor improvement for `LogTag`

nyurik:
* Use standard `From` trait (which is actually only used internally.
* Add Clone & Copy trait

A bit of bike-shedding:
* TBD: usually these things are called `level`, or `log level` - a bit confusing about the tag, but no biggie.
* `Any` is kinda weird tbh, maybe call it `Other(u32)` ?


gquintard:
this isn't bikeshedding, tag are tags, not levels: https://docs.varnish-software.com/tutorials/vsl-query/#varnishlog

fell free to rename `Any()` to `Other()`, no opinion on this

PR #72, opened 2024-09-14
`Event` cleanup

nyurik:
* Add `try_into` to parse `ffi::vcl_event_e` value without panicking
* In backend `wrap_event` callback, panic with a better message including the backend type (is this ok?)
* move all remaining tests to the separate `#[cfg(test)]` mods

### DIFF
--- a/src/vcl/backend.rs
+++ b/src/vcl/backend.rs
@@ -320,7 +320,9 @@ unsafe extern "C" fn vfp_pull<T: Transfer>(
 
 unsafe extern "C" fn wrap_event<S: Serve<T>, T: Transfer>(be: VCLBackendPtr, ev: ffi::vcl_event_e) {
     let backend: &S = validate_director(be).get_backend();
-    backend.event(Event::new(ev));
+    backend.event(ev.try_into().unwrap_or_else(|e| {
+        panic!("{e}: value={ev} for backend {}", backend.get_type());
+    }));
 }
 
 unsafe extern "C" fn wrap_list<S: Serve<T>, T: Transfer>(
--- a/src/vcl/ctx.rs
+++ b/src/vcl/ctx.rs
@@ -191,12 +191,6 @@ impl TestCtx {
     }
 }
 
-#[test]
-fn ctx_test() {
-    let mut test_ctx = TestCtx::new(100);
-    test_ctx.ctx();
-}
-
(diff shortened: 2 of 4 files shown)


gquintard:
does it make sense to use `try_into`? if given `ffi::vcl_event_e`, we must have a corresponding `rust` type for it, that shouldn't fail, ever

nyurik:
sure, but the question is **who** raises the error.  Panic is a very "exception-like" concept that is not common in Rust because it is hard to catch / process / avoid lock poisoning, etc. Instead, it is usually better for any error condition to properly return via `Result<T, E>`, and there is only one place somewhere to decide how an error should be reported to Varnish - i.e. via a panic, or possibly dumping some error file, etc.  That's why I think we should slowly phase out any `unwrap` and `panic`, and instead switch towards `Result` everywhere

gquintard:
it's going to be a lot of tedious work for very little gain. If the C layer hands you a NULL pointer when it shouldn't, or a struct with the wrong magic, you should explode right here and there. Returning a `result` mean the abstraction layer then needs to look at the error and determine whether the error was based on wrong input/avoidable context, or if there's a bug somewhere.

It's really akin to what https://doc.rust-lang.org/1.81.0/src/core/ub_checks.rs.html does when it blows up for https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html . It's a bug to pass it the wrong input, and there's no recovering from it

nyurik:
sure, makes sense.  I actually think that we will refactor events even further once we switch to proc macros -- this value will not be exposed to the users. Something like this:

```rust
#[event(load, unload)]
fn on_load_unload_event() {
  // do something magical
}

#[event(warm)]
fn on_warm_event() {
  // even more magic
}
```

gquintard:
that can work, it'll be a bit annoying to repeat the function signatures, compared to a single function, but, eh.

nyurik:
You can still do it with a single signature, but in that case you do need to have an `enum EventType`, possibly with `Other(::std::ffi::c_uint)`?  Note that `c_uint` is a messy type - it could even be 16bit on some weird systems.

```rust
#[event(all)]
fn on_event(event) {
}
```

nyurik:
That said, I am really not sure if there is ever a case you may want to have a single function for all cases?  Regardless, lets hash it out in an issue on how to do proper design for the events.

PR #75, opened 2024-09-19
Add justfile and use it in workflows

nyurik:
This allows to use the same commands during development and in CI

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,74 @@
+#!/usr/bin/env just --justfile
+
+@_default:
+    just --list --unsorted
+
+# Clean all build artifacts
+clean:
+    cargo clean
+    rm -f Cargo.lock
+
+# Update dependencies, including breaking changes
+update:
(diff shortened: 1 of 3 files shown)


justfile
```diff
@@ -0,0 +1,74 @@
...
+    just --list --unsorted
+
+# Clean all build artifacts
+clean:
+    cargo clean
+    rm -f Cargo.lock
+
+# Update dependencies, including breaking changes
+update:
+    cargo +nightly -Z unstable-options update --breaking
```
gquintard, line 13: do we really have to use nightly here?

nyurik: this is a manual developer-only tool to update dependencies versions. It hasn't been stabilized and could break in the future - which is not a big deal because we don't use it in any automation or builds.  This is just simpler than doing version bumps by hand

gquintard: ok, I was worried it would silently raise the minimal `rust` version needed to compile the project, if it isn't the case, I'll merge today

PR #80, opened 2024-09-25
fail() should not return anything

nyurik:
it seems `fail() -> u8` was a hack to benefit the `vmod_error` example.  `fail()` should not have any results.

Also, a minor unrelated fix - inlined format args in the test.


gquintard:
good catch

PR #81, opened 2024-09-26
Merge 3rd: Proc macro cleanup

nyurik:
A follow up to #54 and #85, and should only be merged after that one (looking at diff doesn't make sense unless you look at the latest commit only, not the whole PR):
* remove VPriv
* switch errors to use `VclError` instead of `Box<Error>` -- more work will be needed here to make them even more optimal
* cleanup all conversion traits. Now just have 3 groups:  `Default<TVcl>`, `From<TVcl> for TRust`, and `IntoVCL<TVcl> for TRust`. The last one is the only custom one, using workspace.
* Wrap many binding-generated types to allow `Default`, `From`, and other messy stuff


nyurik:

PR #82, opened 2024-09-26
Remove warning as error flags for dev

nyurik:
* It is better to allow warnings to stay as warnings while development, but fail warnings in CI
* Also, discovered a bug that I flagged without fixing, just so it won't be forgotten


varnish-sys/src/vcl/convert.rs
```diff
@@ -131,6 +131,8 @@ impl IntoVCL<VCL_STRING> for &[u8] {
     fn into_vcl(self, ws: &mut WS) -> Result<VCL_STRING, String> {
         // try to save some work if the buffer is already in the workspace
         // and if it ends in a null byte
+        // FIXME: UB here - we check if the value AFTER the slice is a null byte
+        //        in other words we access memory that is not ours. This is a bug.
```
gquintard, line 135: whoopsie, indeed

PR #84, opened 2024-09-26
Improve tests, adding glob support

nyurik:
Use proper strings and full paths to run unit tests. Ensures that all test files are executed, rather than forcing devs to remember to add a new .vtc file to the list. Fill fail if no files are found. Also allows users to do their own testing using function calls.

```rust
varnish::run_vtc_tests!("tests/*.vtc");
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ varnish-sys = { path = "./varnish-sys", version = "=0.0.19" }
 #
 # These dependencies are used by one or more crates, and easier to maintain in one place.
 bindgen = "0.70.1"
+glob = "0.3.1"
 pkg-config = "0.3.30"
 serde = { version = "1", features = ["derive"] }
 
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -18,6 +18,7 @@ varnish-sys = { path = "./varnish-sys", version = "=0.0.19" }
 #
 # These dependencies are used by one or more crates, and easier to maintain in one place.
 bindgen = "0.70.1"
+glob = "0.3.1"
```
gquintard, line 21: do we also need it here? Asking as you only seem to be using it in `varnish/`

nyurik: many of these are only used once per some crate.  The goal here is to have all "versioned" crates in one place - so we don't need to look at multiple places to do the version updates -- our crates simply point to this one list.  We also only have one lock file. Also, if one of the dependencies starts being needed in more than one crate, we don't need to move things around - just add another `foo.workspace = true`.

nyurik: P.S. the macros PR adds like 10 more to this list btw

PR #85, opened 2024-09-26
Merge 2nd: Migrate all examples to proc macro

nyurik:
Do not merge until #54 is merged.  To see what has changed in this one, use `commits` tab and click on the second commit.

This PR was part of #54, but now is done separately to separate macro implementation from example migration.  This PR can later be referenced in how old code can be migrated.


nyurik:

PR #86, opened 2024-09-27
Document macro-based approach, bump to v0.1.0

nyurik:
* Add lots of documentation on how to use the new macro-based approach
* This ensures the main branch is always conforming with the semver rules - if the API is breaking, it must increment minor (for 0.*.*), and major (for 1+.*.*).  See https://github.com/obi1kenobi/cargo-semver-checks


gquintard:
versioning a a vanity topic at that point, I'm fine with the checks, but not with upgrading to `0.2.0`, what we have isn't solid enough for that

nyurik:
Massively updated this PR with all the new docs, changelog, etc

PR #89, opened 2024-10-03
Merge 4th: Improvements for other vmods

nyurik:
This PR goes on top of the #81, and adds default handling for different types and `Option<String>` returns.

This change is required for https://github.com/gquintard/vmod-reqwest/pull/13


varnish-macros/src/parser_args.rs
```diff
@@ -180,34 +179,44 @@ impl ParamType {
...
+            let default = Self::get_arg_opts(pat_ty, arg_ty)?;
+            let has_required = Self::get_required_attr(pat_ty)?;
+            let opt = if has_required {
+                if !opt {
+                    error! { "The `required` attribute is only allowed on Option<...> arguments" }
                 }
-                let dflt = Self::get_default(pat_ty, arg_type)?;
-                Self::Value(ParamInfo::new(arg_type, dflt, false))
+                if !arg_ty.must_be_optional() {
+                    error! { "The `required` attribute is only allowed on Probe, ProbeCow, and SocketAddr arguments" }
```
gquintard, line 200: that one is confusing to me, why wouldn't it be allowed on numbers or strings?

nyurik: because it makes no sense to declare an argument as both optional and required unless it is nullable:

```rust
fn foo(#[required] param: Option<i32>) {
    ...
}
```

In theory this might be ok for `&str`, but currently we have a weird duality for strings - so I am not handling it just yet -- i.e. `#[required]` with strings can simply be expressed as `param: &str`.  In the future, once we only allow `Option<&str>`, this list will grow

gquintard: ok, so it would make sense for `&str` :-) I'm good, thanks for the explanation

nyurik:

PR #99, opened 2024-10-28
Cleanup generated enums

nyurik:
* as part of binder build, auto-rename all enum values into Rust values to make them usable and consistent in the user code. For example, `vcl_event_e::VCL_EVENT_LOAD` is now `VclEvent::Load`, and it does NOT need a wrapper enum.
* All enums are tagged as "non-exhaustive" because we do not want to promise that the list of enums will always stay stable (i.e. they can be modified upstream) - so it shouldn't be a breaking change for us.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ varnish-sys = { path = "./varnish-sys", version = "=0.0.19" }
 #
 # These dependencies are used by one or more crates, and easier to maintain in one place.
 bindgen = "0.70.1"
+convert_case = "0.6.0"
 darling = "0.20.10"
 glob = "0.3.1"
 insta = "1"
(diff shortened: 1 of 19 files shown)


gquintard:
I love this, buuuuuuuuuut I would not allow them to be non-exhaustive, I want to see the break if it happens

nyurik:
@gquintard sure - we can mark some as exhaustive and some as not - depending on if we handle them in the wrapper, or if we simply convert them in case some user may want to use it.  The main issue with non-exhaustive ones is  [semver compatibility on enums](https://doc.rust-lang.org/cargo/reference/semver.html#enum-variant-new):

> It is a breaking change to add a new enum variant if the enum does not use the [#[non_exhaustive]](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) attribute.

We do not want to release 0.2.0 just because `VslTag` gained a new enum value.

nyurik:
Moreover, eventually I hope we will stop exporting FFI bindings by default - unless the user enables `ffi` feature.  Obviously this shouldn't happen until we are happy with the Rust-safe API to cover the majority of Varnish functionality.

PR #101, opened 2024-10-29
Some more naming cleanup

nyurik:
Do a few more renames to keep things easier for new users - even I had to lookup these names despite working with Varnish for a while to understand what they do.


gquintard:
I'm not super excited about that one as it would make it harder for C devs coming in to find their marks, but I'm willing to try it

I assume the vmod PRs are coming? ;-)

PR #103, opened 2024-10-30
Name cleanups

nyurik:
* `COWProbe` struct to `CowProbe`
* `COWRequest` struct to `CowRequest`
* `HTTP` struct to `HttpHeaders`
* `HTTPIter` struct to `HttpHeadersIter`
* `VDPCtx` struct to `DeliveryProcCtx`
* `VDP` trait to `DeliveryProcessor`
* `VFPCtx` struct to `FetchProcCtx`
* `VFP` trait to `FetchProcessor`
* `VSB` struct to `Buffer`
* `VSB::cat` function to `Buffer::write`


gquintard:
can we just name VSB "Buffer"? we have context from the namespace already, the scalable part is a bit uncessary?

nyurik:
Makes sense, renamed

PR #105, opened 2024-10-30
Some tiny clippy lint fixes

nyurik:
Many fixes were suggested by clippy (some are no longer listed in the Cargo.toml).  Other than cosmetic changes, did a few minor improvements:

* Add `Clone, Copy` traits to `StreamClose` enum
* `HttpHeaders` now has `iter()` method used by `IntoIterator` implementation. I also removed the IntoIterator on `&mut` - it does not seem to be needed.
* I am still really worried about `transmute::<u32, VslTag>((self.raw.logtag as u32) + u32::from(HDR_FIRST))` -- seems like a major hack, but need feedback if this logic was correct.
* The `ws.alloc` API returns a pointer to the u8, which could be mis-aligned for bigger objects, and is currently marked as "safe". Need to rethink it.
* The `Buffer::write` returns `Result<(), ()>` -- we should return proper error - any thoughts?
* Should `Transfer` trait have an `is_empty()` method?  Clippy suggests it when there is a `len()` method.

### DIFF
--- a/examples/vmod_object/src/lib.rs
+++ b/examples/vmod_object/src/lib.rs
@@ -43,7 +43,8 @@ mod object {
                 .lock() // lock the mutex to access the hashmap
                 .unwrap() // panic if unlocking went wrong
                 .get(key) // look for key
-                .map_or(String::new(), |v| v.to_string()) // used EMPTY_STRING if key isn't found
+                .cloned() // create a copy of the value
+                .unwrap_or_default() // used EMPTY_STRING if key isn't found
         }
 
         /// Insert a key-value pair into the store.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,35 +43,16 @@ similar.opt-level = 3
 unused_qualifications = "warn"
 
 [workspace.lints.clippy]
-# regular warnings - these should probably be fixed first
+# FIXME: safety doc lints are disabled for now, but should be enabled eventually
 missing_safety_doc = "allow"
-new_without_default = "allow"
-len_without_is_empty = "allow"
-not_unsafe_ptr_arg_deref = "allow"  # this one seems pretty bad
-result_unit_err = "allow"
 
 # Pedantic lints - these are more subjective, and some might stay disabled permanently
 pedantic = { level = "warn", priority = -1 }
-bool_to_int_with_if = "allow"
-cast_lossless = "allow"
 cast_possible_truncation = "allow"
 cast_possible_wrap = "allow"
-cast_ptr_alignment = "allow"  # this one is really suspicious
 cast_sign_loss = "allow"
-checked_conversions = "allow"
-derive_partial_eq_without_eq = "allow"
-doc_markdown = "allow"
-implicit_hasher = "allow"
-into_iter_without_iter = "allow"
 missing_errors_doc = "allow"
 missing_panics_doc = "allow"
 module_name_repetitions = "allow"
 must_use_candidate = "allow"
-needless_pass_by_value = "allow"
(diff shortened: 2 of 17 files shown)


gquintard:
> HttpHeaders now has iter() method used by IntoIterator implementation. I also removed the IntoIterator on &mut - it does not seem to be needed.

pretty useful to iterate and modify headers though

> I am still really worried about transmute::<u32, VslTag>((self.raw.logtag as u32) + u32::from(HDR_FIRST)) -- seems like a major hack, but need feedback if this logic was correct.

the tag is an enum and is used as an index, I think there's an easy way around it

> The ws.alloc API returns a pointer to the u8, which could be mis-aligned for bigger objects, and is currently marked as "safe". Need to rethink it.

see my other comment, I don't believe alignment is an issue

> The Buffer::write returns Result<(), ()> -- we should return proper error - any thoughts?

the only error we should be able to it is a lack of space for fixed-size buffers, and that's about it. For dynamic buffers I don't think `realloc()` is an option.

> Should Transfer trait have an is_empty() method? Clippy suggests it when there is a len() method.

meh, if it pleases the linter...

nyurik:
Thx for in depth feedback!  All these were more like notes for further work, as none of them prevent this PR from merging.

> > HttpHeaders now has iter() method used by IntoIterator implementation. I also removed the IntoIterator on &mut - it does not seem to be needed.
> 
> pretty useful to iterate and modify headers though
> 
Sure, but this is NOT what you had in your code https://github.com/gquintard/varnish-rs/blob/f2353ab5f25f2bbd333fc5e98b52ebbe9ec6c795/varnish-sys/src/vcl/http.rs#L225-L228 -- your `next()` function always returned a read-only reference.  We should revisit it later, but for now your API is readonly - so lets keep the code consistent and not introduce too many things in a linting PR :)

> > I am still really worried about transmute::<u32, VslTag>((self.raw.logtag as u32) + u32::from(HDR_FIRST)) -- seems like a major hack, but need feedback if this logic was correct.
> 
> the tag is an enum and is used as an index, I think there's an easy way around it
> 
Yeah, this was more of a red flag than a real issue - lets also table this until later (i didn't make any changes related to it here)

> > The ws.alloc API returns a pointer to the u8, which could be mis-aligned for bigger objects, and is currently marked as "safe". Need to rethink it.
> 
> see my other comment, I don't believe alignment is an issue
> 
Alignment is NOT an issue, the issue is that it returns a pointer to u8 - which should not be cast to a higher alignment ones - i.e. i think the alloc function should already return the needed pointer (this is again a separate PR, not related for this one)

> > The Buffer::write returns Result<(), ()> -- we should return proper error - any thoughts?
> 
> the only error we should be able to it is a lack of space for fixed-size buffers, and that's about it. For dynamic buffers I don't think `realloc()` is an option.

Yep, but we still should uniformly use the same error that can be easily passed up.  The `()` is not an error.

> 
> > Should Transfer trait have an is_empty() method? Clippy suggests it when there is a len() method.
> 
> meh, if it pleases the linter...

you are right, not worth it just yet :)

examples/vmod_object/src/lib.rs
```diff
@@ -43,7 +43,8 @@ mod object {
                 .lock() // lock the mutex to access the hashmap
                 .unwrap() // panic if unlocking went wrong
                 .get(key) // look for key
-                .map_or(String::new(), |v| v.to_string()) // used EMPTY_STRING if key isn't found
+                .cloned() // create a copy of the value
```
gquintard, line 46: duh!

PR #106, opened 2024-10-30
Use much faster DashMap with vmod_kv example

nyurik:
A few followups to this:
* `pub struct` should NOT be allowed inside the vmod - it should be defined outside of it. Not sure why it is currently allowed.
* We may need to rethink `Send`/`Sync` requirements for objects.  Technically, the ownership of the object is sent to Varnish ... which I guess is like another thread - so must be `Send`?

### DIFF
--- a/examples/vmod_object/Cargo.toml
+++ b/examples/vmod_object/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 
 [dependencies]
 varnish.workspace = true
+dashmap = "6.1.0"
 
 [lib]
 crate-type = ["cdylib"]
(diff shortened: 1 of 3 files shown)


gquintard:
I'm pretty reluctant adding a new dependency here unless it's proven to make things better, either speed wise (i.e. it may not matter it the data structure is twice as fast if the time spent on it is dwarfed by the rest of the processing), or CPU wise

gquintard:
> Pointing to the crate that has been downloaded over 80 million times... well, its a good practice :)
something something `right-align` :-)

but, yeah ok, the crate is solid and I get the educational argument

PR #127, opened 2024-11-14
Use `VclError` in WS, CStr support

### LINKED ISSUE #104: Naming is hard, need help
@gquintard I need your naming skills for the few leftover items.... but make sure your Rust hat is on... not the C one :)

* `VSC` -> `Statistics` (?)
* `VSCInternal` -> `StatsInt`  /  `StatsImpl` / ...?
* `VSCBuilder` -> ?
* `VSCBulder::build` -> `attach` -- I think build is a bit confusing here (funny - in this case I actually am siding with the internal Varnish naming)
* `vsm_error` -> ?

This is a followup from #39

nyurik:
- All `Workspace` functions now return `VclError` rather than error strings
- User functions can return `&CStr` as a return value or a `&'static CStr` error
- Fixed test result regeneration
- Mark `Workspace::alloc` as unsafe because the allocated array is not
  initialized
- Rename `VSC` &rarr; `Stats`, `VSCBuilder` &rarr; `StatsBuilder`, and `VSCInternal` into `StatsImpl`
- This breaks published API, so has to be 0.2.0

Fixes #104


varnish-sys/src/vcl/ws.rs
```diff
@@ -78,6 +88,9 @@ impl<'a> Workspace<'a> {
 
     /// Check if a byte slice is allocated in the workspace
     pub fn is_slice_allocated(&self, slice: &[u8]) -> bool {
+        // FIXME: why check for slice size + 1?  In case of a string,
```
gquintard, line 91: Looking at it now. I'm thing we can just check whether the pointer points to an address inside the allocated area of the workspace and call it a day

varnish-sys/src/vcl/ws.rs
```diff
@@ -45,29 +46,38 @@ impl<'a> Workspace<'a> {
     }
 
     /// Allocate a `[u8; sz]` and return a reference to it.
+    ///
+    /// # Safety
+    /// Allocated memory is not initialized with zeroes
```
gquintard, line 51: So we want a safe version that memsets the allocated memory?

nyurik: no, zero-ed memory is not the same - if you have a struct with a constructor that sets one of the fields to a magic number, creating that struct with zeroes is still incorrect.  Instead, we could introduce one or more functions in Workspace like thees:

```rust
fn alloc_default<T: Default>(&mut self) -> &mut T { ... }
fn alloc_from<T>(&mut self, instance: T) -> &mut T { ... }
```

varnish-sys/src/vcl/ws.rs
nyurik: sure, that could also work

PR #129, opened 2024-11-17
rework ws alloc API

nyurik:
- All `Workspace` functions now return `VclError` rather than error strings
- `alloc` was marked as `unsafe`
- `alloc` now returns a `*mut c_void` instead of a `VclResult<&'a mut [u8]>`
- `alloc` now panics if requested size is 0
- Renamed `is_slice_allocated` to `contains`
- Added allocation and copying functions:
  - `allocate(...) -> &mut [MaybeUninit<u8>]` (un-initialized memory)
  - `allocate_zeroed(...) -> &mut [u8]` (initialized memory)
  - `copy_blob(...) -> VCL_BLOB`
  - `copy_txt(...) -> txt`
- Removed `copy_bytes` - use `copy_*` and `alloc_*` functions instead
- Removed `copy_str` - use `copy_cstr` instead


varnish-sys/src/vcl/ws.rs
```diff
@@ -41,105 +45,129 @@ impl<'a> Workspace<'a> {
...
         }
     }
 
-    /// Copy any `AsRef<[u8]>` into the workspace
-    pub fn copy_bytes<T: AsRef<[u8]>>(&mut self, src: &T) -> Result<&'a [u8], VclError> {
-        let buf = src.as_ref();
-        let l = buf.len();
+    /// Allocate memory on Workspace, and move a value into it.
+    /// The value will be dropped in case of out of memory error.
+    pub fn alloc_value<T>(&mut self, value: T) -> Result<&'a mut T, VclError> {
```
gquintard, line 106: I think there's a problem here: the workspace will be wiped at the end of the task, and `rust` will never get the chance to run `drop()`, will it?

varnish-sys/src/vcl/ws.rs
```diff
@@ -41,105 +45,129 @@ impl<'a> Workspace<'a> {
...
-        let l = buf.len();
-        let dest = unsafe { self.alloc(l + 1)? };
-        dest[..l].copy_from_slice(buf);
-        dest[l] = b'\0';
-        // Safe because there are no NULLs in the source, and we just added a NULL
-        Ok(unsafe { CStr::from_bytes_with_nul_unchecked(dest) })
+    /// Copy any `AsRef<[u8]>` into a new [`VCL_BLOB`] stored in the workspace
+    pub fn copy_blob(&mut self, value: impl AsRef<[u8]>) -> Result<VCL_BLOB, VclError> {
+        let buf = self.copy_bytes(value)?;
+        let blob = self.alloc_value(vrt_blob {
```
gquintard, line 130: Performance-wise, we don't care, but it feels a bit silly to create a struct on the stack to copy it, would we be better creating it in the workspace then populating it directly?

varnish-sys/src/vcl/ws.rs
nyurik: You mean `T::drop()` ?  This is correct - but workspace does not have any callback cleanup mechanism, right? And if so, we essentially only want "un-cleanable" types to be copied into workspace.  If so, we may need to mark this as unsafe too.  Current usage of this method is to allocate space for `vrt_blob` and `vrt_backend_probe` 

P.S. should this method be called `copy_value`?

varnish-sys/src/vcl/ws.rs
nyurik: how do you see it?  Let's say we have a slightly different API:

```rust
// Since we don't have any way of dropping T,
// we should use MaybeUninit<T> at first,
// and later switch to ManualDrop<T>...
// not that it matters because we never returned owned objects, only &mut T
fn alloc_type<T>(&mut self) -> VclResult<&mut MaybeUninit<T>>;

// usage
let val = ws.alloc_type::<vrt_blob>();
val.write( vrt_blob { ... } );  // this results in pretty much the same as what I have
```

Alternatively, we can have uninitiized API:

```rust
// result is UB because T is kinda "owned" by self, but hasn't been initialized yet
unsafe fn alloc_type<T>(&mut self) -> VclResult<&mut T>;

// usage
let val = ws.alloc_type::<vrt_blob>();
val.blob = ptr::from_ref(buf).cast::<c_void>();
val.len = buf.len();
val.type_ = Default::default();
```

In short, given that we only deal with a 3 field (24 byte) value, you either copy these 24 bytes 8 byte at a time and eat a cost of dereferencing on each assignment, or copy all 24 bytes at once in one operation...  And now we are deep in the realm of compiler optimization - how smart will LLVM+Rust be to make this much simpler...  I doubt it will play a role.

varnish-sys/src/vcl/ws.rs
gquintard: I think I would get rid of it actually, and only allow to create `&[u8]`, anything else is unsafe, and possibly somewhat useless at the moment

varnish-sys/src/vcl/ws.rs
gquintard: fair, let's optimize for human readability. Ignore my comment!

varnish-sys/src/vcl/ws.rs
nyurik: Agreed, I will make it private (i use it in a few places internally) and rename it to `copy_value`

nyurik:
This is ready to go, but I do have one (non-blocking) question - if someone asks for zero-sized allocation, should this be treated as an error or as a panic?  Out of memory is clearly an error - because we can't predict if there will be enough, plus it can be handled in some ways.  Zero seems like a bug in the code, so panic seems more appropriate... Does that make sense?

gquintard:
the C code will panic if asked to allocate 0 bytes, so you can let it flow to the the native code, or be nice and intercept it to return an error. I don't have a strong opinion, but I agree that it's a bug in the code, so the default panic makes sense.

nyurik:
I am being nice - I require Rust code to call `alloc(NonZeroUsize)` -- so Rust code must construct NonZeroUsize object - which by definition requires the user to verify that the value is non-zero

PR #130, opened 2024-11-17
Make Probe and CowProbe into one struct

nyurik:
Keeping it much simpler - the only real difference between them is `String` vs `Cow<str>` for two fields.  A simple generic fixes that.


gquintard:
Looks good to me, I'll merge one CI passes

nyurik:
thx, passed

PR #137, opened 2024-11-19
Implement safe filter support

### LINKED ISSUE #135: Improve filter management
This is a proposed implementation of the filter registration process. It shows just the fetch filter (`vfp`), but the `vdp` is nearly identical (too bad we can't seem to have just one implementation other than with a macro to de-dup the code).

### New VCL state

When handling `#[shared_per_vcl]` values of user type `T`,  and storing `Box(T)` directly in the `vp: *mut vmod_priv`, we get a new type `PerVclState<T>`.  If `vmod_priv` is NULL, we create and store a default one.

```rust
#[derive(Debug, Default)]
pub struct PerVclState<T> {
    pub fetch_filters: Vec<Box<ffi::vfp>>,
    pub delivery_filters: Vec<Box<ffi::vdp>>,
    pub user_data: Option<Box<T>>,
}
```

### User's event method

In order to register a new filter, user can now ask for a special filter registry parameters

```rust
pub fn event(event: Event, vfp: &mut FetchFilters, vdp: &mut DeliveryFilters) {
    if let Event::Load = event {
        // user type implements FetchProcessor trait
        fetch_filters.register::<UserFetchFilterType>();
         // user type implements DeliveryProcessor trait
        delivery_filters.register::<UserDeliveryFilterType>();
    }
}
```

### FetchFilters implementation

This struct gets temporarily created just for the duration of the user's function. Note that both `FetchFilters` and `DeliveryFilters` hold a `&mut` into the `PerVclState` struct - i.e. using partial exclusive borrowing.  The ctx ref is readonly, so both can share the same ctx ref.  We may have to store `*const vrt_ctx` instead if we run into the borrowchecker issues.

```rust
#[derive(Debug)]
pub struct FetchFilters<'c, 'f> {
    ctx: &'c vrt_ctx,
    filters: &'f mut Vec<Box<ffi::vfp>>,
}

impl<'c, 'f> FetchFilters<'c, 'f> {
    fn new(ctx: &'c vrt_ctx, filters: &'f mut Vec<Box<ffi::vfp>>) -> Self {
        Self { ctx, filters }
    }

    fn find_position<T: FetchProcessor>(&self) -> Option<usize> {
        let name = T::name().as_ptr();
        self.filters.iter().position(|f| f.name == name)
    }

    pub fn register<T: FetchProcessor>(&mut self) -> bool {
        if self.find_position::<T>().is_none() {
            let instance = Box::new(new_vfp::<T>());
            unsafe {
                ffi::VRT_AddVFP(self.ctx, instance.as_ref());
            }
            self.filters.push(instance);
            true
        } else {
            false
        }
    }

    pub fn unregister<T: FetchProcessor>(&mut self) -> bool {
        if let Some(pos) = self.find_position::<T>() {
            let filter = self.filters.swap_remove(pos);
            unsafe {
                ffi::VRT_RemoveVFP(self.ctx, filter.as_ref());
            }
            true
        } else {
            false
        }
    }

    pub fn unregister_all(&mut self) {
        for filter in self.filters.drain(..) {
            unsafe { ffi::VRT_RemoveVFP(self.ctx, filter.as_ref()) }
        }
    }
}
```

### Cleanup

Macro generator will create a static cleanup thing like this, and `on_fini` will unregister everything

```rust
static PRIV_VCL_METHODS: vmod_priv_methods = vmod_priv_methods {
    magic: VMOD_PRIV_METHODS_MAGIC,
    type_: c"...".as_ptr(),
    fini: Some(vmod_priv::on_fini::<PerVclState<...>>),
};
```

nyurik:
Fixes #135


gquintard:
it looks good, and I'll merge if you are done with it, but heads-up: I need to rework the filters a bit, which may mean we register actual values, rather than letting the `register` function create them, but that will come later

PR #138, opened 2024-11-19
Update crate metadata

nyurik:


Cargo.toml
```diff
@@ -6,11 +6,15 @@ members = ["varnish", "varnish-macros", "varnish-sys", "vmod_test", "examples/vm
 [workspace.package]
 # These fields are used by multiple crates, so it's defined here.
 # Version must also be updated in the `varnish-macros` dependency below.
-version = "0.2.0"
+version = "0.2.1"
```
gquintard, line 9: We haven't published 0.2.0 yet, we don't really need to bump the revision just yet, do we?

gquintard: Ah, nevermind you just did

PR #146, opened 2024-11-22
6.0 support

gquintard:
@nyurik , I'm going to need an adult to handle the sharp macros/features :-)

Right now, I've gated everything I could in `varnish-sys` using the `_lts_60` feature and `cargo b` works at the root, but things blow up in `generator.rs` wen trying to compile `vmod-example`:

```
error[E0432]: unresolved import `varnish`
  --> examples/vmod_example/src/lib.rs:7:1
   |
7  | #[varnish::vmod(docs = "README.md")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `VMOD_PRIV_METHODS_MAGIC` in `ffi`
   |
```

it makes sense, `ffi` doesn't expose that item in 6.0, but the feature is in `varnish-sys`, not in `varnish-macro`.

Also, note that I've tweaked the `justfile` and `Dockerfile` to actually install `varnish` 6.0, previously it was relying on the highest version, meaning the official 6.6.


docker/Dockerfile
```diff
@@ -17,8 +17,12 @@ RUN    apt-get update \
 
 # Pass in the VARNISH_VERSION build arg to specify the version of Varnish to install
 ARG VARNISH_VERSION=76
-RUN curl -sSf "https://packagecloud.io/install/repositories/varnishcache/varnish${VARNISH_VERSION}/script.deb.sh" | sudo bash \
-    && (sudo apt-get install -y varnish-dev || sudo apt-get install -y libvarnishapi-dev)
+RUN set -ex; \
+    curl -sSf "https://packagecloud.io/install/repositories/varnishcache/varnish${VARNISH_VERSION}/script.deb.sh" | sudo bash; \
+    echo 'Package: varnish varnish-dev\nPin: origin "packagecloud.io"\nPin-Priority: 1001' > /etc/apt/preferences.d/varnish; \
```
nyurik, line 22: I guess we will need some sort of an `if` statement here to dynamically decide how to install each version - unless you think this will work for all versions?

varnish-sys/build.rs
```diff
@@ -112,6 +113,9 @@ fn find_include_dir(out_path: &PathBuf) -> Option<(Vec<PathBuf>, String)> {
             if major == 7 && minor <= 5 {
                 println!("cargo::rustc-cfg=feature=\"_objcore_in_init\"");
             }
+            if major == 6 && minor <= 5 {
```
nyurik, line 116: should the `minor` be checked here?  Seems like it should only test for `major == 6` ?

varnish-sys/src/validate.rs
```diff
@@ -1,10 +1,15 @@
...
+    director, req, sess, vrt_ctx, ws, DIRECTOR_MAGIC, REQ_MAGIC,
+    SESS_MAGIC, VCL_BACKEND, VRT_CTX_MAGIC, WS_MAGIC,
 };
+#[cfg(not(feature = "_lts_60"))]
+use crate::ffi::{vcldir, vfp_ctx, vfp_entry, VCLDIR_MAGIC, VFP_CTX_MAGIC, VFP_ENTRY_MAGIC};
+
+#[cfg(not(feature = "_lts_60"))]
 use crate::vcl::{DeliveryFilters, FetchFilters};
 
+#[cfg(not(feature = "_lts_60"))]
```
nyurik, line 12: i guess we should move a lot of these to a new file to simplify gating

nyurik (commented):
wow, that's quiet a few gates...

docker/Dockerfile
gquintard: this will work for all packages coming from packagecloud, which is what we want to use

varnish-sys/build.rs
gquintard: copy-pasta, we really only want to support `minor == 0` here

varnish-sys/src/validate.rs
gquintard: I haven't look at that yet, but yes, that would be nice

varnish-macros/src/generator.rs
```diff
@@ -157,7 +164,11 @@ impl Generator {
         }
 
         let json = serde_json::to_string_pretty(&json! {json}).unwrap();
-        format!("VMOD_JSON_SPEC\u{2}\n{json}\n\u{3}")
+        if cfg!(lts_60) {
+            format!("{json}")
```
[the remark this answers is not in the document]
gquintard: shhhhhhhhhhh

varnish-sys/build.rs
```diff
@@ -112,6 +115,9 @@ fn find_include_dir(out_path: &PathBuf) -> Option<(Vec<PathBuf>, String)> {
             if major < 6 || major > 7 {
                 println!("cargo::warning=Varnish v{ver} is not supported and may not work with this crate");
             }
+            if major == 6 && minor == 0 {
```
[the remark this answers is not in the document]
gquintard: nah, I really don't want people on anything other than 6.0 or the 2 freshest versions

varnish-macros/src/generator.rs
```diff
@@ -188,12 +199,12 @@ impl Generator {
...
             VCL_VOID,
             VMOD_ABI_Version,
-            VMOD_PRIV_METHODS_MAGIC,
             VclEvent,
             vmod_priv,
-            vmod_priv_methods,
             vrt_ctx,
         ];
+        #[cfg(not(lts_60))]
+        use_ffi_items.append_all(["VMOD_PRIV_METHODS_MAGIC,", "vmod_priv_methods,"]);
```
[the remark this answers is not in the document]
gquintard: yup, realized that later on
point taken about the non-trivialitym but, does it really matter?

varnish-macros/src/generator.rs
```diff
@@ -93,13 +93,20 @@ impl Generator {
...
-               type_: #ty_name.as_ptr(),
-               fini: Some(vmod_priv::#on_fini::<#ty_ident>),
-           };
-        });
+        if cfg!(lts_60) {
+            tokens.push(quote! {
+                static #name: vmod_priv_free = Some(vmod_priv::#on_fini::<#ty_ident>);
+            });
+        } else {
+            #[cfg(not(lts_60))]
```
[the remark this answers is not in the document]
gquintard: no, but I feel like you're shooting fish in a barrel, reviewing an super early draft :-p

varnish-sys/build.rs
```diff
@@ -112,6 +113,9 @@ fn find_include_dir(out_path: &PathBuf) -> Option<(Vec<PathBuf>, String)> {
             if major < 6 || major > 7 {
                 println!("cargo::warning=Varnish v{ver} is not supported and may not work with this crate");
             }
+            if major == 6 && minor == 5 {
+                println!("cargo::rustc-cfg=feature=\"_lts_60\"");
```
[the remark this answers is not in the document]
gquintard: corrected during our call, that makes things a lot clearer

gquintard:
have at it @nyurik :-)

I'm unsure on how to group all the gates, if you have tweaks, feel free to propose, or to impose (just commit to this branch, but keep the history for now)

nyurik:
1) I wonder if we should stop useing `lts_60` style conditionals, but instead follow the way browsers do it -- via feature detection.  For example, if `vpriv` cleanup should be done with a struct, declare `varnishapi_vpriv_struct_cleanup` - and depending on that, either generate a `fini` struct with a method, or pass just the function.  The added bonus of this approach is that we will not have a million `lts60` gates, but instead will be able to see what relates just to the vpriv cleanup codegen.

2) I don't think we should focus on LTS - instead, it might be easier to just support 6.0 (and add it to testing), but don't explicitly call it `lts` - because this way we MAY eventually be able to support other 6.0  **IF** someone wants to maintain support for it. In other words, we choose to support specifically 6.0, and test for it, but we don't have to declare it as having a special name.

3) please help with the varnishlib installation - its having some issues in Github CI runner, and package management is not my strong side

4) testing is obviously a bit tricky here. We could do several things:
* for compiler tests, we can use some convention, e.g. if a test filename starts with `v6`, only run it if some global cfg flag is declared
* for examples, we may need some other mechanism to prevent the test from even running - either place it in a separate `examples-v6` dir, or some other method. Unsure yet.

gquintard:
1. I'm not sure I catch your drift here, would you have a short code example to demonstrate the idea?
2. sure, as long as we have a recognizable name we can search for, it's just naming
3. will look at that
4. it should just be shell scripting in `just` and `ci`, shouldn't it?

PR #147, opened 2024-11-22
Easier build config reuse

nyurik:
* Use `links` manifest key (see [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key)) - this exports all metadata fields to all dependants.

With this change, any crate that depends on `varnish-sys` will have `DEP_VARNISHAPI_VERSION_NUMBER=6.0.13` env var set during the build, accessible from that crate's `build.rs`.  We can re-parse it, or better yet - add some more metadata values relevant to specific usecases.

* Do not declare hidden features in `Cargo.toml` - turns out there is a [special command](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-check-cfg) for that.
* Added a few tools to the docker

### DIFF
--- a/varnish-sys/build.rs
+++ b/varnish-sys/build.rs
@@ -1,4 +1,5 @@
 use std::path::PathBuf;
+use std::str::Split;
 use std::{env, fs};
 
 use bindgen_helpers::{rename_enum, Renamer};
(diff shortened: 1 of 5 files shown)


varnish-sys/build.rs
```diff
@@ -7,6 +8,9 @@ static BINDINGS_FILE: &str = "bindings.for-docs";
 static BINDINGS_FILE_VER: &str = "7.6.1";
 
 fn main() {
+    // <=7.5 passed *objcore in vdp_init_f as the 4th param
+    println!("cargo::rustc-check-cfg=cfg(varnishsys_objcore_in_init)");
```
gquintard, line 12: that should be conditional, shouldn't it?

gquintard: oh, nevermind, I should learn to read

gquintard (approved):

PR #155, opened 2024-12-08
No type

gquintard:
Here's an idea, trying to make the `Backend` API a bit more palatable. Main items:
- no more `get_type()` method in the `Serve` trait, just stash it at creation
- create a `handle` field wrapping the pointer so you at least can't destroy the backend as easily while returning a dangling pointer
- the vmod code can return the handle and the conversion to `VCL_BACKEND` is done through macros
- `Serve::get_headers()` now returns `Body`, which can be empty, `AsRef<[u8]>` or a `Read`

I like that API quite a bit, you can notably check the updated examples in `backend.rs` for the `Read` case, and `vmod_be` for the `AsRef<[u8]>`, but there are a couple of caveats:
- using `dyn` to avoid specializing the functions means the vmod write needs to box objects, which I'm not a fan of, but that can be alleviated using helper functions
- main problem is that right `vmod_be` clones the data, and it'd be so much nicer to just use a reference, but I ran out of time before finding a way to not impose lifetime hell down to the writer. A quick solution would be ti use `Arc`, but there's probably something cleaner?


gquintard:
I'm still not happy about it because it uses an `Arc`, uses more boilerplate and we really don't need it (because we are real programmers who know what they are doing /s) but at lease we don't copy the `Vec`:

```
struct Sentence {
    data: VecWrap,
}  
  
#[derive(Clone)]  
struct VecWrap (Arc<Vec<u8>>);
  
impl AsRef<[u8]> for VecWrap {  
    fn as_ref(&self) -> &[u8] {  
        &*self.0  
    }  
}  
  
// a lot of the Serve trait's methods are optional, but we do need to implement  
// get_headers() that sets the response headers, and returns a Body  
impl Serve for Sentence {  
    fn get_headers(&self, ctx: &mut Ctx) -> Result<Body, VclError> {  
        let beresp = ctx.http_beresp.as_mut().unwrap();  
        beresp.set_status(200);  
        beresp.set_header("server", "parrot")?;  
  
        Ok(Body::Buffer(Box::new(self.data.clone())))
    }  
}  
```

nyurik:
i just merged-in main branch, resolving some conflicts

nyurik:
Also, i reverted the generation of the API.md - this is a complicated PR, and API.md gets in the way :)

gquintard:
OBE. Hopefully we can build something on top of #246

PR #161, opened 2025-01-10
Clippy 1.84 fixes

nyurik:


gquintard (approved):
This will break my backend PR (it remove the `get_type()` method, so consider this a reminder to have a look, but please do merge this :-)

PR #163, opened 2025-01-12
add stats macros

mnutt:
This adds macros for a vmod to expose stats (counters, gauges) into varnish. For example:

```rust
#[varnish::stats]
pub struct Stats {
    // Number of hits
    #[counter]
    hits: std::sync::atomic::AtomicU64,

    // The temperature outside
    //
    // This tends to vary from one location to the next.
    #[gauge(level = "debug")]
    temperature: std::sync::atomic::AtomicU64,
}

let stats = Stats::new("mystats", "default");
stats.hits.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```

Then you would be able to see it in `varnishstat`:

```
$ varnishstat -1 |grep mystats
mystats.default.hits                       1         0.04 Number of hits
mystats.default.temperature                0         0.00 The temperature outside
```

## Implementation

This works by defining a struct to hold the counters. Varnish itself will allocate the memory to hold the struct, and then in rust we cast the memory to the struct type and fill it with values. This means we need to be careful and only hold a reference to the struct rather than copying it; I ended up with generating an inner struct from what the user supplied and creating an outer struct with `Deref` to point to the inner struct.

This currently generates the counter definition JSON by string interpolation; if we're ok with `json_serde` as a dependency we could use that instead.


nyurik (commented):
I am really happy this is being worked on!  Note that I don't have that much inner knowledge about the counters themselves, so I might be misunderstanding some things.

I left a few inline comments for the internal things, plus here's a more general thoughts:

* I am not yet sold on the idea this should be a proc macro - the only code that is per-struct dependent is the list of field names in the generated `get_struct_metrics`. The rest of the generated code like `new` should be some library function with a generic constraint, not generated code.  Instead of all this, perhaps create a new trait like `VscMetric`, and use `#[derive(VscMetric)]` to generate the need function that returns some field name metadata.
* As I mentioned in a chat -- generating code is a lot of fun to do, but is a very bad "first" step -- because we have not yet agreed on **what** needs to be generated.  If we decide that `VscMetric` trait is needed (per above) - you can simply implement it by hand as part of the example - so that we can see clean rust code rather than using `cargo expand` and try to analyze what it generates.
* the lifetime of the stats object is really confusing - what destroys it? Who calls your custom `drop` fn?  And why not rely on the standard `Drop` trait?
* I think this is a bit cleaner API (still work in progress);

```rust
use std::ops::Deref;

#[derive(VscMetric)]
#[repr(C)] // required, checked by VscMetric
pub struct UserStats {
    // fields are declared as you did before
}

struct obj {
    stats: Vsc<UserStats>,
}

#[varnish::vmod]
mod my_vmod {
    impl obj {
        pub fn new(ctx: &mut Ctx) -> Self {
            Self {
                obj: Vsc::new(ctx, UserStats {})
            }
        }
    }
}

// API

/// trait is unsafe because it requires #[repr(C)] on the struct,
/// and requires extensive validation of the fields.
/// The derive macro will generate the required metadata and do the validation.
unsafe trait VscMetric {
    /// returns a string with the metadata of the struct as internal Varnish JSON.
    /// users won't know how to generate this correctly, but might be ok at first.
    /// in the future, we may want to return a struct with the metadata, so that users can customize it
    fn get_metadata() -> &'static str;
}

struct Vsc<T: VscMetric> {
    // ...
}

impl<T: VscMetric> Vsc<T> {
    /// creates a new Vsc object with the given stats object 
    pub fn new(ctx: &mut Ctx, stats: T) -> Vsc<T> {
        // ... VRT_VSC_Alloc ...
    }
}

impl Drop for Vsc<T> {
    fn drop(&mut self) {
        // ... VRT_VSC_Destroy ...
    }
}

impl Deref for Vsc<T> {
    type Target = T;
    fn deref(&self) -> &Self::Target {
        // ...
    }
}
```


See also [repr(C) discussion](https://users.rust-lang.org/t/does-repr-c-define-a-trait-i-can-use-to-check-structs-were-declared-with-repr-c/16323/6)

mnutt:
Thanks for the review, and that revised API looks good--I suspected that my solution occupied an uncomfortable middle between "just use your own decorated rust struct" and a DSL like `vsctool.py`. It looked like it was giving a copy of the struct you defined, but isn't really and is liable to break expectations.

Regarding `vsc_seg`, the `drop()` method not being part of a Drop impl was just an oversight on my part.

It may take me a day or two to have another revision up.

gquintard:
Alright, I promised @nyurik a review and a bit more context of what's happening in the C world, so here we go.

Varnish counters are essentially `mmap`'d files, describing the counter structures they embed, so that external processes can discover and read them.

All starts with a `VSC` file, an example of this would be https://github.com/varnishcache/varnish-cache/blob/master/vmod/VSC_debug.vsc:
```
..
	Copyright (c) 2021 Varnish Software AS
	SPDX-License-Identifier: BSD-2-Clause
	See LICENSE file for full text of license

#
.. varnish_vsc_begin::	debug
	:oneliner:	Example Counters
	:order:		70

	Test counters from vmod_debug

.. varnish_vsc:: count
	:type:		counter
	:level:		debug
	:oneliner:	arbitrary counter

	Just counting something

.. varnish_vsc_end::	debug
```

which allows the build system to generate two files. First is `vmod/VSC_debug.h`:
```
struct VSC_debug {
	uint64_t	count;
};

#define VSC_debug_size PRNDUP(sizeof(struct VSC_debug))

struct vsmw_cluster;
struct vsc_seg;

struct VSC_debug *VSC_debug_New(struct vsmw_cluster *,
    struct vsc_seg **, const char *fmt, ...);
void VSC_debug_Destroy(struct vsc_seg **);
```

i.e. we get a constructor, a descructor and the struct definition. Notably, @mnutt, I messed up: the struct that you create in rust but match this, so we can't use `AtomicU64`, they have to be `uint64_t` and the struct must be `repr(C)`.

Updating the struct will update the file and the tools reading it will be automatically updated. If you want to play with the reader code, you can check https://github.com/gquintard/varnish-counters-demo

The second file `vmod/VSC_debug.c`:
```
#include "config.h"
#include <stdio.h>
#include <stdarg.h>
#include "vdef.h"
#include "vas.h"
#include "vrt.h"
#include "VSC_debug.h"

static const char vsc_debug_name[] = "DEBUG";

#define PARANOIA(a,n)				\
    v_static_assert(				\
	offsetof(struct VSC_debug, a) == n,	\
	"VSC element '" #a "' at wrong offset")

PARANOIA(count, 0);
#undef PARANOIA

static const unsigned char vsc_debug_json[307] = {
	123,34,118,101,114,115,105,111,110,34,58,34,49,34,44,34,110,97,109,101,
	34,58,34,100,101,98,117,103,34,44,34,111,110,101,108,105,110,101,114,
	34,58,34,69,120,97,109,112,108,101,32,67,111,117,110,116,101,114,115,
	34,44,34,111,114,100,101,114,34,58,55,48,44,34,100,111,99,115,34,58,
	34,92,116,84,101,115,116,32,99,111,117,110,116,101,114,115,32,102,114,
	111,109,32,118,109,111,100,95,100,101,98,117,103,34,44,34,101,108,101,
	109,101,110,116,115,34,58,49,44,34,101,108,101,109,34,58,123,34,99,111,
	117,110,116,34,58,123,34,116,121,112,101,34,58,34,99,111,117,110,116,
	101,114,34,44,34,99,116,121,112,101,34,58,34,117,105,110,116,54,52,95,
	116,34,44,34,108,101,118,101,108,34,58,34,100,101,98,117,103,34,44,34,
	111,110,101,108,105,110,101,114,34,58,34,97,114,98,105,116,114,97,114,
	121,32,99,111,117,110,116,101,114,34,44,34,102,111,114,109,97,116,34,
	58,34,105,110,116,101,103,101,114,34,44,34,105,110,100,101,120,34,58,
	48,44,34,110,97,109,101,34,58,34,99,111,117,110,116,34,44,34,100,111,
	99,115,34,58,34,92,116,74,117,115,116,32,99,111,117,110,116,105,110,
	103,32,115,111,109,101,116,104,105,110,103,34,125,125,125,0
};

// {
//   "version": "1",
//   "name": "debug",
//   "oneliner": "Example Counters",
//   "order": 70,
//   "docs": "\tTest counters from vmod_debug",
//   "elements": 1,
//   "elem": {
//     "count": {
//       "type": "counter",
//       "ctype": "uint64_t",
//       "level": "debug",
//       "oneliner": "arbitrary counter",
//       "format": "integer",
//       "index": 0,
//       "name": "count",
//       "docs": "\tJust counting something"
//     }
//   }
// }


struct VSC_debug*
VSC_debug_New(struct vsmw_cluster *vc,
    struct vsc_seg **sg, const char *fmt, ...)
{
	va_list ap;
	struct VSC_debug *retval;

	va_start(ap, fmt);
	retval = VRT_VSC_Alloc(vc, sg, vsc_debug_name, VSC_debug_size,
	    vsc_debug_json, sizeof vsc_debug_json, fmt, ap);
	va_end(ap);
	return(retval);
}

void
VSC_debug_Destroy(struct vsc_seg **sg)
{
	struct vsc_seg *p;

	AN(sg);
	p = *sg;
	*sg = NULL;
	VRT_VSC_Destroy(vsc_debug_name, p);
}
```

As you can see, the constructor and destructor are very simple, most of the work lies in creating the JSON description

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    hits: std::sync::atomic::AtomicU64,
+
+    /// Temperature in degrees Celsius
+    #[gauge]
+    temperature: std::sync::atomic::AtomicU64,
+
+    /// Memory usage in bytes
+    ///
+    /// Memory usage can vary quite a bit, based on the number of foo objects.
+    #[gauge(level = "debug", format = "bytes")]
```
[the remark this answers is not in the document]
gquintard: yes, it can't dynamically change once a counter has been created. You could imagine an API where every counter can be created with a different level, but I don't really see a benefit here, setting it in the struct definition seems good enough

nyurik:
Atimicu64 is actually a u64, see https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html

varnish/src/vsc_types.rs
```diff
@@ -0,0 +1,68 @@
+use varnish_sys::ffi::vsc_seg;
+
+pub struct VscMetricDef<'a> {
```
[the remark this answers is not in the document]
gquintard: I have a different take: those don't need to be public. The reader code will have access to them, using the JSON document, but this code updates the counter and knows what they are, so no need to expose those

gquintard:
> Atimicu64 is actually a u64, see https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html

fair, but isn't it confusing to not use `u64`?

nyurik:
if the functionality of a type must be restricted, we shouldn't use the "raw" type.  u64 allows much more operations than atomicu64 (bit shifts, increments, ...) - all of which should not be done to a counter.  Also, atomic allows internal mutability via a read-only reference, whereas u64 must be mutable to modify.  In our case, we may even want to introduce one more layer on top of the atomic type, exposing only the relevant functionality.

gquintard:
using a new type works for me, I was just arguing against using `AtomicU64`

nyurik:
it has to be atomicu64 to allow internal mutability and correct expectations (including not-compiling it for platforms that don't support atomic u64s. You can always wrap it in yet another type, but tbh i don't see much value in it just yet

gquintard:
Ah, I see, that makes sense now, thank you.

Aside from that, I agree with the review comment that overloading types is a bad idea, so I'd like to spell out an idea and see if it doable/sane, and I have one  question about implementation too.

Essentially, I'd just like an `Exposable` trait that the macro implements, and and `Exposed` trait that flows from here. So possibly something like this:

```
// it would possibly make sense to make this trait implementable only by the macro to avoid
// users writing garbage JSON for incompatible structures, but I'm unsure how to do that
trait VCSJSON {
    fn json_metadata() -> &'static str;
}

// the stats field is private, so the type can't be built manually by vmod writers
struct ExposedStats<T: Exposable> {
    stats: T,
}

Regarding types, I
impl ExposedStat<T> {
    pub fn new() -> Self,
}

// ExposedStats<T> also implements Drop, Deref for &T 
```

The main question I have here is: can we prevent users from implementing `VCSJSON` wrong?

nyurik:
I must be missing something - why not simply keep the existing design - only allow atomicu64, tag it with some attributes, and generate json internally by the macro (either directly, or by having some sort of a description struct, whose instances would generate correct json).  This is what i proposed with the `fn get_metadata() -> &'static str;` in my review above.  You don't need to expose json generation to the user

gquintard:
if we just augment the struct instead of wrapping it, doesn't it make it trivial for the user to wrongly make it `Copy`? Or even to manually create that struct?

If we don't use `VRT_VSC_Alloc()`, the counters won't be exposed

nyurik:
You can't make the struct Copy or Clone because `AtomicU64` does not implement those traits.

You can instantiate your own struct, sure, but creating it without registering is pointless.  It is obviously possible for the user to create some random struct and forget to register it with Varnish, but I don't think this is a very common failure case - as oppose to having really convoluted and opaque code generation.

gquintard:
Ah, good call out on the `AtomicU64`, I forgot about that.

But, then, how do we automatically deregister the function?

nyurik:
if the user struct will implement our custom trait, the user will register a new struct instance -- which will take ownership and return a wrapper.  The wrapper will unregister on drop

nyurik:
ah, right, it is nearly identical, thx. My only concern is the `pub fn new() -> Self,` - i don't think it should be public -- the wrapper constructor should be internal, created by the registration function

gquintard:
I was actually thinking that `ExposedStats<MyStats>::new()` would both create a new `MyStats`, with all fields being zeroed, build the wrapper and register `MyStats`

nyurik:
that's the design in my first comment at the top, something like this:

```rust
// Instantiating the wrapper type, passing in the custom user type
// Note that the user type should be instantiated by the user
// We don't need to know how to do it, i.e. we don't need to require `default` trait
impl<T: VscMetric> Vsc<T> {
    pub fn new(ctx: &mut Ctx, stats: T) -> Vsc<T> {
        // this is actually an error:
        // &mut ctx cannot be both inside Vsc as well as be used to call VRT_VSC_Alloc
        // this means Vsc would have to store an unsafe pointer instead of a &mut ref
        let v = Vsc { stats, ctx };
        // some registration process
        ctx.VRT_VSC_Alloc(&v);
        // user gets back ownership of the wrapper type, storing it in the shared obj
        v
    }
}

impl Drop for Vsc<T> {
    fn drop(&mut self) {
        // object already contains a pointer to ctx which it uses to unregister itself
        self.ctx.VRT_VSC_Destroy(&self);
    }
}
```

mnutt:
I think this should bring it in line with the suggested changes above. I'll bet there might be some naming suggestions though..

I wasn't sure about the context allocating and destroying the objects? (`ctx.VRT_VSC_Alloc(&v)`) As far as I can tell, the wrapper gets managed by the context, and when the wrapper goes out of scope, `drop` deallocates the counter?

examples/vmod_counters/src/lib.rs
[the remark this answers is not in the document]
mnutt: oh I missed that one--better to set it on the struct instead of on the attribute? I was imagining the vmod developer might just have a single counter struct for their whole vmod, and might have some mix of debug, info, and diag. I guess they could always make separate structs for that if they wanted?

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,65 @@
...
+            VRT_VSC_Destroy(self.name.as_ptr(), self.seg);
+        }
+    }
+}
+
+impl<T: VscMetric> Deref for Vsc<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { &*self.metric }
```
nyurik, line 57: I guess rust will ensure that `deref` only gets called before any `Drop`, so this should be safe

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+impl<T: VscMetric> DerefMut for Vsc<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { &mut *self.metric }
+    }
+}
```
nyurik, line 65:
```suggestion
}

```

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,65 @@
...
+    metric: *mut T,
+    seg: *mut vsc_seg,
+    name: CString,
+}
+
+impl<T: VscMetric> Vsc<T> {
+    pub fn new(module_name: &str, module_prefix: &str) -> Self {
+        let mut seg = std::ptr::null_mut();
+        let name = CString::new(module_name).unwrap();
+        let format = CString::new(module_prefix).unwrap();
```
nyurik, line 20: use `expect("...")` to explain why these two could fail... i am guessing neither should be coming from the Varnish users, but only from the vmod authors, otherwise we would need to change this to `-> Result<Self, ...>` - so that all the error cases get properly returned.

varnish-macros/src/lib.rs
```diff
@@ -64,3 +66,33 @@ pub fn vmod(args: pm::TokenStream, input: pm::TokenStream) -> pm::TokenStream {
...
+/// Handle the `#[derive(Stats)]` macro. This can only be applied to a struct.
+/// The struct must have only fields of type `AtomicU64`.
+/// - `#[counter]` attribute on a field will export it as a counter.
+/// - `#[gauge]` attribute on a field will export it as a gauge.
+#[proc_macro_derive(Stats, attributes(counter, gauge))]
+pub fn get_metadata(input: pm::TokenStream) -> pm::TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+    
+    if !stats::has_repr_c(&input) {
```
nyurik, line 79: the content of this function should be moved to the `stats.rs` -- this way `proc_macro::TokenStream` (aliased as `pm`) is only used by the compiler to call this function, but the content uses `proc_macro2::*` types -- which allows testing of all this functionality.  Just keep the `parse_macro_input!` here

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+use varnish::vsc_wrapper::Vsc;
+use varnish::Stats;
+
+#[derive(Stats)]
+#[repr(C)] // required for correct memory layout
+pub struct VariousStats {
+    /// Number of hits
+    #[counter]
+    hits: std::sync::atomic::AtomicU64,
```
nyurik, line 9:
```suggestion
    hits: AtomicU64,
```

and add `use std::sync::atomic::AtomicU64` at the top

examples/vmod_counters/src/lib.rs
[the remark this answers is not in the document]
nyurik: I don't know what's the best usecase here. I don't see any problems with multiple counters if needed. I am still a bit unclear how the level affects the code and if the level should be set on each field or the whole struct - i'll let @gquintard guide this one

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+use varnish::vsc_wrapper::Vsc;
+use varnish::Stats;
+
+#[derive(Stats)]
+#[repr(C)] // required for correct memory layout
+pub struct VariousStats {
+    /// Number of hits
```
nyurik, line 7: perhaps spell out what "hit" means?  Doc that repeats field name is not that useful :)

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+pub struct Vsc<T: VscMetric> {
+    metric: *mut T,
+    seg: *mut vsc_seg,
+    name: CString,
+}
+
+impl<T: VscMetric> Vsc<T> {
+    pub fn new(module_name: &str, module_prefix: &str) -> Self {
```
nyurik, line 17: Do we need any kind of context pointer here?  Or does `VRT_VSC_Alloc` rely on a global singleton?

nyurik (commented):
awesome progress!  Some minor fixes (see comments).  Also, we really need some testing - similar to `varnish-macros/src/tests.rs` -- create a new dir `varnish/tests/pass_vsc`, and add a simple file like this:

```
#[derive(Stats)]
#[repr(C)]
pub struct Stats1 {
    #[counter]
    count: AtomicU64,
}
```

modify `run_parse_tests` to also check for structs that have `derive(Stats)` - and create a new function `test_vsc` similar to `test` -- it would take a `TokenStream` for the struct, and generate the `vsc_code` output -- basically calling the same internal function as your `get_metadata` (see my comment on `get_metadata` itself)

and the test function

gquintard:
> I wasn't sure about the context allocating and destroying the objects?

I wouldn't. The context/object owns the stat struct already, so the lifetime is already properly handled

examples/vmod_counters/src/lib.rs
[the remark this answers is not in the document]
gquintard: @mnutt , I think you are good here, the level is per-field and is fixed upon the struct creation, so using an attribute the way you did makes complete sense to me

varnish/src/lib.rs
```diff
@@ -99,10 +99,12 @@ pub mod ffi {
 pub use varnish_sys::ffi;
 
 pub mod varnishtest;
+pub mod vsc_wrapper;
 
 #[cfg(feature = "vsc")]
 pub mod vsc;
 
+pub use varnish_macros::Stats;
```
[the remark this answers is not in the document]
gquintard: I disagree. `vsc` is for reading counters from external programs, I think thecode in this PR should go into `vmod::counters` or `vmod:vsc`, but in any case, we want to keep the current `vsc` module, even if we rename is `VSCReader` or something to avoid confusion

varnish/src/vsc_wrapper.rs
gquintard: singleton (`heritage` in the C code)

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,65 @@
...
+        let mut seg = std::ptr::null_mut();
+        let name = CString::new(module_name).unwrap();
+        let format = CString::new(module_prefix).unwrap();
+        
+        let metadata_json = T::get_metadata();
+
+        let metric = unsafe {
+            VRT_VSC_Alloc(
```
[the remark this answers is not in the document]
gquintard: yes, memory is created for us and we only own the wrapper, and we won't get notified if the counter set is destroyed, but Varnish won't do it behind our backs

examples/vmod_counters/tests/test01.vtc
```diff
@@ -0,0 +1,39 @@
...
+	import stats from "${vmod}";
+
+	sub vcl_init {
+		new c = stats.test();
+	}
+
+	sub vcl_deliver {
+		c.increment_hits();
+		c.increment_hits();
+		c.update_temperature(17);
```
gquintard, line 18: can we make it a function of a header to change it more than once?

examples/vmod_counters/tests/test01.vtc
```diff
@@ -0,0 +1,39 @@
...
+# Send another request to verify increment
+client c2 {
+	txreq
+	rxresp
+	expect resp.http.x-hits == "4"
+} -run
+
+# Assert on the VSC counter value
+varnish v1 -expect mystats.default.hits == 4
+varnish v1 -expect mystats.default.temperature == 17
```
gquintard, line 39: can you load a new VCL in `v1` after all is done so we can make sure the counters have disappeared?

varnish/src/vsc_wrapper.rs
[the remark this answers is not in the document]
nyurik: why aren't we getting notified?  The code below implements wrapper's `Drop` - does that mean we must call `VRT_VSC_Destroy` - or else it will result in a leak?

we need to prevent the user from doing something like this:
* instantiate VSC wrapper, e.g. as part of object's new method
* save it to some global context, e.g. like this:

```rust
// global
thread_local! {
    static STORED_OBJECT: RefCell<Option<Wrapper>> = RefCell::new(None);
}
fn set_object(value: Wrapper) {
    STORED_OBJECT.with(|obj| { *obj.borrow_mut() = Some(value); });
}
```

* access the wrapper later - using some method that internally dereferences a pointer -- undefined behavior here

Which means we must have some way for the user-owned wrapper object to either have a way to check pointer validity, or get notified when the pointer is no longer valid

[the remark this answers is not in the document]
gquintard: as long the wrapper exists, `VRT_VSC_Destroy` isn't called, so we are good.
My point is: if `VRT_VSC_Destroy` was called by somebody else, we wouldn't know, but nobody but us will call it

examples/vmod_counters/src/lib.rs
mnutt: I can make something up 😄 I was struggling to come up with something plausible-seeming that was simple enough that it didn't raise unnecessary red herrings

gquintard: honestly, I don't reckon you need to go for something realistic. I'd be fine with something like:
```
/// Some arbitrary counter
foo: AtomicU64,
```

and in the vmod, have a `.add_one_foo()` function

varnish/src/lib.rs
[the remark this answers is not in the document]
mnutt: makes sense to me, happy to move it to whereever you'd like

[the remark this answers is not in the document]
gquintard: `varnish:vmod:Stats` sounds good to me?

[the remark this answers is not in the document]
nyurik: vmod is not a namespace, its an attribute.  I would actually move it to the root - `varnish::Stats`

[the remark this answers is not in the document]
gquintard: ah right, we changed that, `varnish::Stats` it is then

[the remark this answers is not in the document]
nyurik: that said, `Stats` is an incredibly generic word :D... Plus this is a name of a trait, making it even more confusing.  Lets do some fun bike shading?

* `VarnishCounter`
* `CounterStats`
* `VclCounter`
* `VclStatistics`  <-- I personally like this one, but not hard set
* ...?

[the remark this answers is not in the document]
gquintard: `CounterStats` sounds highly redundant.

Isn't it enough to have them in `varnish::vcl::Stats`? `VclCounter` and `VclStatistics` contains a "Vcl` that is basically implied by that path.

Plus, using `varnish::vcl` means the counter-reading code can live in `varnish::vsc`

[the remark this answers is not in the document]
nyurik: it is a bit redundant , agree, but consider how it is used:

```rust
use varnish::VclStatistics;

#[derive(VclStatistics)]
struct MyStats { /* ... */ }
```


```rust
use varnish::vcl::Statistics;

#[derive(Statistics)]
struct MyStats { /* ... */ }
```


```rust
use varnish::vcl;  // this is not used very often, but doable

#[derive(vcl::Statistics)]
struct MyStats { /* ... */ }
```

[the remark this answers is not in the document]
gquintard: I don't have strong feelings, but I still prefer `Statistics`, repeating `Vcl` "everywhere" (in practice you will see it in the `use`, `derive` and in the struct definition) seems unnecessary

As long as we avoid "counters", I'm ok

varnish-macros/src/lib.rs
nyurik: this is still a todo - we have to have a way of testing the output of the proc macro

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+use serde::Serialize;
+use syn::{punctuated::Punctuated, token::Comma, Data, Field, Fields, Type};
+
+type FieldList = Punctuated<Field, Comma>;
+
+#[derive(Serialize, Clone)]
+struct VscMetricDef {
+    pub name: String,
+    #[serde(rename = "type")]
+    pub counter_type: String, // "counter", "gauge"
```
nyurik, line 11: `counter_type` should be an enum. Also, perhaps call it `metric_type` for readability?

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub ctype: String,    // "uint64_t" is only option right now
```
nyurik, line 12: I think we may want to introduce enums for `ctype` and `level`.  Strings are not great for validation

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+fn generate_metrics(fields: &FieldList) -> Vec<VscMetricDef> {
+    fields
+        .iter()
+        .enumerate()
+        .map(|(i, field)| {
+            let name = field.ident.as_ref().unwrap().to_string();
+
+            let counter_type = if field
```
nyurik, line 68: Let's create a utility function like `has_attr` and put it next to `remove_attr` in the `parser_utils.rs`.  TIL you can remove certain attributes automatically by adding them to the `proc_macro_derive` as a second parameter - I had to remove them because i didn't know it existed

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+
+            VscMetricDef {
+                name,
+                counter_type: counter_type.to_string(),
+                ctype: "uint64_t".to_string(),
+                level,
+                oneliner,
+                format,
+                docs,
+                index: Some(i * 8),
```
nyurik, line 95: need a comment to explain why `8` (i'm guessing that's the size of `uint64_t` - but this seems inherently wrong because we shouldn't just multiply, but rather track how big is the structure so far, and append the size of the current element to it.  Of course this hack does work for now because all items are the same size

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub level: String,    // "info", "debug", etc
+    pub oneliner: String, // "Counts the number of X", etc
+    pub format: String,   // "integer", "bytes", "duration", "bitmap", etc
+    pub docs: String,
+    pub index: Option<usize>,
+}
+
+#[derive(Serialize)]
+struct VscMetadata {
+    version: String,
```
nyurik, line 22: If its always a constant, no need to allocate

```suggestion
    version: &'static str,
```

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+}
+
+pub fn generate_metadata_json(name: &str, fields: &FieldList) -> String {
+    let metrics = generate_metrics(fields);
+
+    let metadata = VscMetadata {
+        version: "1".to_string(),
+        name: name.to_string(),
+        oneliner: format!("{name} statistics"),
+        order: 100,
```
nyurik, line 108: need to comment the meaning of 100

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,162 @@
...
+        docs: String::new(),
+        elements: metrics.len(),
+        elem: metrics
+            .iter()
+            .map(|m| (m.name.clone(), m.clone()))
+            .collect(),
```
nyurik, line 114: no need to clone.  Even better would be to change `generate_metrics` to return the HashMap - otherwise you need to iterate twice and waste a vector allocation

```suggestion
        elem: metrics.into_iter().map(|m| (m.name.clone(), m)).collect(),
```

nyurik (commented):
Looking better, thanks! I added a few more comments, and there are a few older comments still left to cleanup, but i think this is a good path

gquintard:
alright, I had a try implementing it on [vmod-reqwest](https://github.com/gquintard/vmod-reqwest/tree/stats) and I really love how simple it is. Problem is that all tests explode in `VRT_VSC_Alloc`:
```
**** v1    CLI RX|Signal 11 (Segmentation fault) received at (nil) si_code 1
**** v1    CLI RX|version = varnish-7.6.1 revision c3d5882003eb87e5e93dc09fb9513ca96db3ca3c, vrt api = 20.1
**** v1    CLI RX|ident = Linux,6.12.10-arch1-1,x86_64,-jnone,-sdefault,-sdefault,-hcritbit,epoll
**** v1    CLI RX|now = 6071.921601 (mono), 1739149792.125922 (real)
**** v1    CLI RX|Backtrace:
**** v1    CLI RX|  0x59d8d0547f75: varnishd(VBT_format+0x75) [0x59d8d0547f75]
**** v1    CLI RX|  0x59d8d04b4634: varnishd(+0x54634) [0x59d8d04b4634]
**** v1    CLI RX|  0x59d8d05416be: varnishd(VAS_Fail+0x1e) [0x59d8d05416be]
**** v1    CLI RX|  0x59d8d04a5aa9: varnishd(+0x45aa9) [0x59d8d04a5aa9]
**** v1    CLI RX|  0x73ebc32251d0: /usr/lib/libc.so.6(+0x3d1d0) [0x73ebc32251d0]
**** v1    CLI RX|  0x59d8d055348d: varnishd(VSB_vprintf+0x9d) [0x59d8d055348d]
**** v1    CLI RX|  0x59d8d04e8d3f: varnishd(VSMW_Allocv+0x15f) [0x59d8d04e8d3f]
**** v1    CLI RX|  0x59d8d04e912b: varnishd(+0x8912b) [0x59d8d04e912b]
**** v1    CLI RX|  0x59d8d04e9268: varnishd(VRT_VSC_Alloc+0xa8) [0x59d8d04e9268]
**** v1    CLI RX|  0x73ebb8132551: ./vmod_cache/_vmod_reqwest.8b18a7d3e5830c38d1bfca08fa91ff8ea185838ea4b472a18e42a16d34a29c0f(+0xd8551) [0x73ebb8132551]
**** v1    CLI RX|  0x73ebb813ae33: ./vmod_cache/_vmod_reqwest.8b18a7d3e5830c38d1bfca08fa91ff8ea185838ea4b472a18e42a16d34a29c0f(+0xe0e33) [0x73ebb813ae33]
**** v1    CLI RX|  0x73ebb812a929: ./vmod_cache/_vmod_reqwest.8b18a7d3e5830c38d1bfca08fa91ff8ea185838ea4b472a18e42a16d34a29c0f(+0xd0929) [0x73ebb812a929]
**** v1    CLI RX|  0x73ebb8101681: ./vmod_cache/_vmod_reqwest.8b18a7d3e5830c38d1bfca08fa91ff8ea185838ea4b472a18e42a16d34a29c0f(+0xa7681) [0x73ebb8101681]
**** v1    CLI RX|  0x73ebc030b864: vcl_vcl1.1739149791.648497/vgc.so(VGC_function_vcl_init+0xc4) [0x73ebc030b864]
**** v1    CLI RX|  0x73ebc030e77e: vcl_vcl1.1739149791.648497/vgc.so(+0x777e) [0x73ebc030e77e]
**** v1    CLI RX|  0x59d8d04c6bbb: varnishd(+0x66bbb) [0x59d8d04c6bbb]
**** v1    CLI RX|  0x59d8d04c8c01: varnishd(+0x68c01) [0x59d8d04c8c01]
**** v1    CLI RX|  0x59d8d054b7ab: varnishd(+0xeb7ab) [0x59d8d054b7ab]
**** v1    CLI RX|  0x59d8d054bd5d: varnishd(VCLS_Poll+0x38d) [0x59d8d054bd5d]
**** v1    CLI RX|  0x59d8d048d355: varnishd(CLI_Run+0x75) [0x59d8d048d355]
```

It's probably just some pointer misstep, but I haven't had the time to look at it. I'll try to dig this week.

Some non-urgent notes for my future self:
- we'll want to expose `VRT_VSC_Hide` and `VRT_VSC_Reveal` to help with user readability, notably in the backend cases
- I'm not a fan of `vsc_wrapper`, but `vsc` is already taken. Maybe this should be `vsc` and use `vsc_reader` for the other one
- `bitmap` type will be useful for backend
- personal research: have a look at which ordering makes sense in which conditions

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,67 @@
...
+                std::ptr::null_mut(),
+                &mut seg,
+                name.as_ptr(),
+                size_of::<T>(),
+                metadata_json.as_ptr(),
+                metadata_json.len(),
+                format.as_ptr(),
+                std::ptr::null_mut(),
```
gquintard, line 34:
```suggestion
                varnish_sys::ffi::va_list::default(),
```

if `NULL` things explode on my laptop

varnish-macros/src/stats.rs
mnutt: Hrm, I looked into it and it appears to do...nothing? I don't see it getting read in the varnish-cache source. I guess I can just try leaving it off entirely.

mnutt:
Should have changes in for pretty much everything except the `vsc_wrapper` rename and testing the macro, I was running into some trouble trying to test it.

I also may have to reconsider the HashMap `generate_metrics`, as it doesn't guarantee stable order. Might not matter much in the library itself, but makes it harder to test.

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,89 @@
+use std::sync::atomic::AtomicU64;
+use varnish::vsc_wrapper::Vsc;
+use varnish::Stats;
+
+#[derive(Stats)]
+#[repr(C)] // required for correct memory layout
+pub struct VariousStats {
```
nyurik, line 7: How about `ExampleStats` ?

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Some arbitrary counter
+    #[counter]
+    foo: AtomicU64,
```
nyurik, line 10:
```suggestion
    foo: AtomicU64, // only AtomicU64 are supported at this point
```

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,226 @@
...
+#[derive(Serialize, Clone)]
+#[serde(rename_all = "lowercase")]
+enum Format {
+    Integer,
+    Bitmap,
+    Duration,
+    Bytes,
+}
+
+impl Default for Format {
```
nyurik, line 54: you can do this by simply adding `#[default]` next to `Integer` in the enum, and adding `Default` to the derive attribute

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,67 @@
+use std::ffi::CString;
+use std::mem::size_of;
+use std::ops::{Deref, DerefMut};
+use varnish_sys::ffi::{va_list, vsc_seg, VRT_VSC_Alloc, VRT_VSC_Destroy};
+
+pub unsafe trait VscMetric {
+    fn get_metadata() -> &'static str;
+}
+
+pub struct Vsc<T: VscMetric> {
```
nyurik, line 10: How about `MetricsContainer` instead?  A bit more verbose, but much more readable

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,67 @@
+use std::ffi::CString;
+use std::mem::size_of;
+use std::ops::{Deref, DerefMut};
+use varnish_sys::ffi::{va_list, vsc_seg, VRT_VSC_Alloc, VRT_VSC_Destroy};
+
+pub unsafe trait VscMetric {
```
nyurik, line 6: How about `Metric` or `Metrics` or `MetricSource` or ...?

nyurik (commented):
some naming suggestions.  Also, take a look at my last #166 as it relates to all this a bit

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,226 @@
...
+fn parse_metric_attributes(field: &Field, metric_type: &str) -> (Level, Format) {
+    let mut level = Level::default();
+    let mut format = Format::default();
+
+    if let Some(attrs) = find_attr(&field.attrs, metric_type) {
+        let _ = attrs.parse_nested_meta(|meta| {
+            match meta.path.get_ident().map(ToString::to_string).as_deref() {
+                Some("level") => {
+                    let level_str = meta.value()?.parse::<syn::LitStr>()?.value();
+                    level = match level_str.as_str() {
```
nyurik, line 187: Don't do this here.  Add this next to `enum Level`:

```rust
impl FromStr for Level {
    type Err = ();

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "info" => Ok(Self::Info),
            "diag" => Ok(Self::Diag),
            "debug" => Ok(Self::Debug),
            _ => Err(()),
        }
    }
}
```

varnish-macros/src/stats.rs
```diff
@@ -0,0 +1,226 @@
...
+                        "info" => Level::Info,
+                        "diag" => Level::Diag,
+                        "debug" => Level::Debug,
+                        _ => panic!("Invalid level value for field {}. Must be one of: info, diag, debug", 
+                            field.ident.as_ref().unwrap()),
+                    };
+                }
+                Some("format") => {
+                    let format_str = meta.value()?.parse::<syn::LitStr>()?.value();
+                    format = match format_str.as_str() {
```
nyurik, line 197: same as above.

nyurik (commented):
parsing shouldn't be done inline - we have nice traits for that - which also makes it reusable

varnish/src/vsc_wrapper.rs
```diff
@@ -0,0 +1,67 @@
...
+        let metric = unsafe {
+            VRT_VSC_Alloc(
+                std::ptr::null_mut(),
+                &mut seg,
+                name.as_ptr(),
+                size_of::<T>(),
+                metadata_json.as_ptr(),
+                metadata_json.len(),
+                format.as_ptr(),
+                va_list::default(),
```
gquintard, line 34:
```suggestion
                va_list::default().as_mut_ptr(),
```

sorry, I probably suggested the wrong thing, we need a pointer here

mnutt: Hmm, this is tricky. At least on macos, I don't seem to have `va_list::default()` at all, unless I'm missing something? It seems as if the actual implementation of `va_list` might be compiler-specific.

nyurik:
I tried to rebase, and had to squash the history to make it doable.  I also made a few minor renames.  Note that the test case still fails - needs some @gquintard magic

nyurik:
@gquintard I fixed the clippy thing, but that's not what's failing :'( (I mean - it did fail in CI, but the `just test` fails for some other reason)

gquintard:
thanks, so, yes, that's the issue we've identified around `vargs`: at the moment there's no `rust` implementation that will easily work on all platforms

nyurik:
@gquintard that totally solves them!  Now I can simply include that function in our C wrapper code, compile it as part of the regular build process, and once it becomes available upstream - rely on that instead.  See `VRT_VSC_AllocVariadic` in the code and the new build script.

varnish-macros/src/stats.rs
nyurik: @gquintard any thoughts on this?

examples/vmod_counters/tests/test01.vtc
```diff
@@ -0,0 +1,39 @@
...
+		set resp.http.x-hits = c.get_hits();
+	}
+} -start
+
+client c1 {
+	txreq
+	rxresp
+	expect resp.http.x-hits == "2"
+} -run
+
```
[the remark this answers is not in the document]
gquintard: for example https://github.com/varnishcache/varnish-cache/blob/master/bin/varnishtest/tests/b00003.vtc#L27

varnish-macros/src/metrics.rs
```diff
@@ -0,0 +1,321 @@
...
+    Info,
+    Diag,
+    Debug,
+}
+
+impl FromStr for Level {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        // TODO: do we really need to compare with ignore_ascii_case?
```
gquintard, line 56: I don't thing we do

varnish-macros/src/metrics.rs
```diff
@@ -0,0 +1,321 @@
...
+        // TODO: consider using strum::FromRepr crate instead?
```
gquintard, line 57: let's?

varnish-macros/src/metrics.rs
```diff
@@ -0,0 +1,321 @@
...
+        if s.eq_ignore_ascii_case("info") {
+            Ok(Level::Info)
+        } else if s.eq_ignore_ascii_case("diag") {
+            Ok(Level::Diag)
+        } else if s.eq_ignore_ascii_case("debug") {
+            Ok(Level::Debug)
+        } else {
+            Err(format!(
+                // TODO: can s come from an untrusted source? If so, we should not include it
```
gquintard, line 66: indeed, we should break hard on anything that we don't support here

but, actually, where is this function needed?

examples/vmod_counters/src/lib.rs
```diff
@@ -0,0 +1,89 @@
+use std::sync::atomic::AtomicU64;
+use varnish::vsc_wrapper::Vsc;
+use varnish::Stats;
+
+#[derive(Stats)]
+#[repr(C)] // required for correct memory layout
```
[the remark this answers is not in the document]
nyurik: we won't be able to use a `derive` -- the function with `#[proc_macro_derive]` returns code that gets added after the struct, but it can't alter the struct itself.  We could instead use a more generic `proc_macro` - just like we did for vmods.

```rust
#[varnish_stat]
struct Stats {
  ...
}
```

varnish-macros/src/metrics.rs
nyurik: the problem with the strum::FromRepr is that it would use the same capitalization as the enum itself... and it seems we want all lowercase?

examples/vmod_counters/src/lib.rs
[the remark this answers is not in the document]
gquintard: it's a bit of a footgun, isn't it? What happens if the user derives but forgets the `repr`?

varnish-macros/src/metrics.rs
gquintard: indeed, so we've considered it and decided against, boom, done :-)

examples/vmod_counters/src/lib.rs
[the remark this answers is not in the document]
nyurik: we check for it. we just can't generate it, unless we use proc_macro instead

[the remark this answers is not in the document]
nyurik: usually, derived proc macros are better because they are "scoped" - i.e. the user knows that derive macro can only generate trait implementation(s) and possibly some other extra code, but it can't modify the item itself.  With proc macro, it is a wild west - anything goes, any changes, and it might get too magical...

gquintard:
@nyurik , do you want to tackle the few remaining discussions so we can merge this?

nyurik (approved):

PR #164, opened 2025-01-17
no more Backend::get_type()

gquintard:


gquintard:
@nyurik, the CI is broken because of Varnish packages not installing, which I'll look at, however, running locally, it looks `bindgen` is creating code that upsets `clippy`:
```
$ just ci-test
rustc --version
rustc 1.83.0 (90b35a623 2024-11-26)
cargo --version
cargo 1.83.0 (5ffbef321 2024-10-29)
cargo fmt --all -- --check
cargo clippy --all-targets --workspace $(/home/guillaume/.cargo/bin/just get-package-exclude-args) -- -D warnings
warning: varnish-sys@0.4.0: Generated bindings from Varnish 7.6.1 differ from checked-in bindings.for-docs. Update with   cp /home/guillaume/work/varnish-rs/target/debug/build/varnish-sys-74fd726f6665b6b4/out/bindings.rs varnish-sys/bindings.for-docs
    Checking varnish-sys v0.4.0 (/home/guillaume/work/varnish-rs/varnish-sys)
error: calling `CStr::new` with a byte string literal
   --> /home/guillaume/work/varnish-rs/target/debug/build/varnish-sys-74fd726f6665b6b4/out/bindings.rs:711:14
    |
711 |     unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(b"Log\0") };
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use a `c""` literal: `c"Log"`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_c_str_literals
    = note: `-D clippy::manual-c-str-literals` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::manual_c_str_literals)]`
```

nyurik:
hm, this is weird - i tried `just ci-test` with both 1.83 and 1.84 and both worked. I also tried it in a docker container - also no issues. Please do `just clean` and see if you can reproduce in a docker?

varnish-sys/src/vcl/backend.rs
```diff
@@ -98,8 +96,8 @@ pub struct Backend<S: Serve<T>, T: Transfer> {
     #[expect(dead_code)]
     methods: Box<ffi::vdi_methods>,
     inner: Box<S>,
-    #[expect(dead_code)]
-    type_: CString,
+    #[allow(dead_code)]
```
nyurik, line 99: why allow?  expect is better because it will not compile if not needed

gquintard:
uh, nevermind, that worked

varnish-sys/src/vcl/backend.rs
```diff
@@ -119,11 +117,12 @@ impl<S: Serve<T>, T: Transfer> Backend<S, T> {
     /// Create a new builder, wrapping the `inner` structure (that implements `Serve`),
     /// calling the backend `name`. If the backend has a probe attached to it, set `has_probe` to
     /// true.
-    pub fn new(ctx: &mut Ctx, name: &str, be: S, has_probe: bool) -> VclResult<Self> {
+    pub fn new(ctx: &mut Ctx, type_: &str, name: &str, be: S, has_probe: bool) -> VclResult<Self> {
```
nyurik, line 120: perhaps use `ctype` here too? or give it a more descriptive name (like `backend_id`?)

varnish-sys/src/vcl/backend.rs
```diff
@@ -351,12 +345,33 @@ unsafe extern "C" fn wrap_pipe<S: Serve<T>, T: Transfer>(
     sc_to_ptr(backend.pipe(&mut ctx, tcp_stream))
 }
 
+// CStr is tied to the lifetime of bep, but we only use it for error messages
+unsafe fn get_type(bep: VCL_BACKEND) -> &'static str {
```
nyurik, line 349: why not add this as a method to the `impl VCL_BACKEND` ?

varnish-sys/src/vcl/backend.rs
gquintard: `backend_type`? it should be the same for all backend of a same category

varnish-sys/src/vcl/backend.rs
gquintard: I think I wrote that one before you moved to `expect`, easy to change

varnish-sys/src/vcl/backend.rs
gquintard: that would be smart, wouldn't it?

varnish-sys/src/vcl/backend.rs
nyurik: bikeshedding the name a bit more - `type` here seem to mean "the name of the struct that represents a class of  backends", right? Or is this a specific instance of a backend?  So perhaps `backend_info` or `backend_id` or ... ?  `type` is a bit confusing here...

nyurik: `backend_category`? :)

gquintard: I don't really like going with `id` here as it suggests (to me at least) some uniqueness.
`backend_info` works for me, I'll do that

gquintard: `category` works to

nyurik: how is this different from "backend name" ?

nyurik: in short - do we really need to have two names here?

gquintard: there's already `name` which is the name of the backend as it will appear in `varnishadm backend.list`, and usually, it's the name of the VCL object representing the object.

So you can create multiple backends from the same vmod, and you would expect them to be from the same category but to have different names

nyurik: so one is the name given by the user, i.e. a backend "variable" -- in this case `polly`:

```vcl
new polly = be.parrot("I'll repeat this on and on");
```

and the other is the backend type, i.e. `parrot`?   Although backend and object are technically not the same thing... So anyway, maybe `be_type` and `be_name` as params?

Side thought - seems like we should have called it `#[object_name]` or `#[obj_name]` instead of `#[vcl_name]`... (we could rename it and support both - not difficult)...

gquintard: `be_type`/`be_name` it is

nyurik: would it make sense to call it `backend_id` instead of `backend_name`?  Does "be_name" imply any kind of uniqueness?

gquintard: id feels more numerical? the name is expected to be unique I believe.
I can change it, no strong opinions

nyurik: name? I thought name is arbitrary and ok to dup, whereas id is more of a unique thing -- but I guess i spent far too much time in  higher level langs for an ID to frequently mean a unique string...  perhaps `backend_id` and `backend_type` are the cleanest?  Also, note that people try to spell things out for public apis (i often ignore that, but... i shouldn't :) )

gquintard:
all good?

nyurik (approved):
perfect, thanks! :)

PR #165, opened 2025-01-17
no more 7.4 in CI

gquintard:
`github actions` moved to `ubuntu:noble` and `varnish:7.4` isn't packaged for it, on top of being EOL

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -20,7 +20,6 @@ jobs:
           - setup: varnish76
             type: latest
           - setup: varnish75
-          - setup: varnish74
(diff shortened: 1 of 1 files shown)


nyurik:
if its not crashing or causing any problems, why remove?  I would much rather have more CI test coverage just for information purposes - so we know when something stops working - and remove it then

gquintard:
7.4 isn't packaged for noble, which is now `ubuntu-latest`, which I'd like to use. I don't feel like installing an old ubuntu just to CI an EOL Varnish version

nyurik:
ah, ok

nyurik (approved):

PR #166, opened 2025-02-12
`vsc` module cleanup

nyurik:
A bunch of VSC-related renames per our chat. This way we won't conflict between metrics reader and metrics provider

* Fixed FFI memory bug - `patience` was passing "off" without the null termination to C
* Modified `patience`, refactoring it and making it return `Self` rather than `Result<Self>` 
* deleted `vsc` feature
* `mod vsc` -> inlined at the top crate level
* `Stats` -> `MetricsReader`
* `StatsImpl` -> `MetricsReaderImpl` (private)
* `StatsBuilder` -> `MetricsReaderBuilder`
* `Stat` -> `Metric`
* `Format` -> `MetricsFormat`
* Removed `Format::Unknown` value. This is a very `C-like` pattern that's almost non-existent in Rust. Converted `From` to `TryFrom` -- as it now can return an error  - using `()` for now.

## TODO:
* [ ] Should `Semantics` be renamed into something?
* [ ] `patience` has an internal assert - is this needed? Should it return a real result instead?
* [ ] `patience` uses `self.vsm`, but the `vsc_arg` below uses `self.vsc`?
* [ ] The builder uses `Result<Self, Err>` return pattern - I'm not sure how easy it will be to use... we might want to think if we want `Result<&mut Self, Err>` instead? I'm not certain on this yet

### DIFF
--- a/varnish/src/metrics_reader.rs
+++ b/varnish/src/metrics_reader.rs
@@ -74,17 +74,15 @@ impl<'a> MetricsReaderBuilder<'a> {
     /// it specifies the timeout to use.
     #[must_use]
     pub fn patience(self, t: Option<Duration>) -> Self {
-        let mut arg = match t {
+        let arg = CString::new(match t {
             None => "off".to_string(),
             Some(t) => t.as_secs_f64().to_string(),
-        }
-        .into_bytes();
-        arg.push(0);
+        })
+        .unwrap(); // Can never fail since we control the content of the string
 
         // # Safety
         // we just created this string, no point to double-check it for nul bytes
         unsafe {
-            let arg = CString::from_vec_with_nul_unchecked(arg);
             // TODO: document why this can fail, and if we should return an error
             // TODO: Document why using `self.vsm` here, and `self.vsc` in the other `VSM_Arg` calls
             let ret = ffi::VSM_Arg(self.vsm, 't' as c_char, arg.as_ptr());
--- a/varnish/src/metrics_reader.rs
+++ b/varnish/src/metrics_reader.rs
@@ -230,18 +230,19 @@ pub enum MetricFormat {
     Bitmap,
     /// Time
     Duration,
+    /// Unit unknown
+    // FIXME: This should contain original value
+    Unknown,
 }
 
-impl TryFrom<c_int> for MetricFormat {
-    type Error = ();
-
-    fn try_from(value: c_int) -> Result<Self, Self::Error> {
-        match From::from(u8::try_from(value).map_err(|_| ())?) {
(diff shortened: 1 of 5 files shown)


gquintard (approved):

gquintard (approved):

gquintard:
regarding TODO items:
- no opinion on semantics, I don't really have a better option in mind
- I think the assert in `patience` is justified, it should always work
- VSC is a subset of the VSM, same for VSL. It might be cleaner to just duplicate the VSM arguments into both VSC and (non-existing as of yet) VSL functions
- for the builder, it may make sense to stash the arguments, and just do all the calls in the last step, that would clean the pattern

varnish/src/metrics_reader.rs
```diff
@@ -223,46 +230,48 @@ pub enum Format {
     Bitmap,
     /// Time
     Duration,
-    /// Unit unknown
```
gquintard, line 226: As mentioned, I'm very much against that change, `vmods` can create unknown types, and we can't fail on those...

varnish/src/metrics_reader.rs
```diff
@@ -223,46 +230,48 @@ pub enum Format {
...
+    let stat = Metric {
         value: point.ptr,
         name: CStr::from_ptr(point.name).to_str().unwrap(),
         short_desc: CStr::from_ptr(point.sdesc).to_str().unwrap(),
         long_desc: CStr::from_ptr(point.ldesc).to_str().unwrap(),
         semantics: point.semantics.into(),
-        format: point.format.into(),
+        // FIXME: Unknown formats should continue to work
+        //        Perhaps use `Option<MetricFormat>` for `format` field?
+        format: point.format.try_into().unwrap(),
```
gquintard, line 274: ...like you do here

gquintard (changes requested):

varnish/src/metrics_reader.rs
nyurik: If `unknown` is a valid value, we certainly can keep it - but in that case it cannot be `Unknown`, but rather `Custom(c_int)` or something similar, i.e. it must represent some state, rather than be an uncertain value.

gquintard: I agree, do you want to do it in this PR, or do that later?

gquintard (approved):

varnish/src/metrics_reader.rs
nyurik: separate PR

nyurik: Created https://github.com/varnish-rs/varnish-rs/issues/178 to remember

PR #168, opened 2025-02-12
fix overeager WS caching

### LINKED ISSUE #167: missing checks in IntoVCL to safely avoid work
in `0.3.0`, our the `IntoVCL` implementation is a bit too lazy for `&[u8]`: we only check that the array is in the workspace, but not that it's null-terminated

gquintard:
fixes #167

### DIFF
--- a/varnish-sys/src/vcl/convert.rs
+++ b/varnish-sys/src/vcl/convert.rs
@@ -267,13 +267,19 @@ impl IntoVCL<VCL_STRING> for &[u8] {
     fn into_vcl(self, ws: &mut Workspace) -> Result<VCL_STRING, VclError> {
         // Try to save some work if the buffer is already in the workspace.
         // We assume that &[u8] has always been readonly, so workspace data is valid.
-        if ws.contains(self) {
+        // However, we need to make sure that the buffer is null-terminated, either by its own last
+        // byte, or by the one after it
+        let with_extra_byte = unsafe { std::slice::from_raw_parts(self.as_ptr(), self.len() + 1) };
+        if (ws.contains(self) && self.ends_with(b"\0"))
+            || (ws.contains(with_extra_byte) && with_extra_byte.last() == Some(&b'\0'))
+        {
             Ok(VCL_STRING(self.as_ptr().cast::<c_char>()))
         } else {
             Ok(VCL_STRING(ws.copy_bytes_with_null(self)?.b))
         }
     }
 }
+
 impl IntoVCL<VCL_STRING> for &str {
     fn into_vcl(self, ws: &mut Workspace) -> Result<VCL_STRING, VclError> {
         self.as_bytes().into_vcl(ws)
--- a/varnish/tests/vmod_test/src/lib.rs
+++ b/varnish/tests/vmod_test/src/lib.rs
@@ -16,7 +16,9 @@ mod rustest {
     use std::time::Duration;
 
     use varnish::ffi::VCL_STRING;
-    use varnish::vcl::{CowProbe, Ctx, Event, FetchFilters, Probe, Request, VclError, Workspace};
+    use varnish::vcl::{
+        CowProbe, Ctx, Event, FetchFilters, IntoVCL, Probe, Request, VclError, Workspace,
+    };
 
     use super::VFPTest;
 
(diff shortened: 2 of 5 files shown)


varnish-sys/src/vcl/convert.rs
```diff
@@ -267,13 +267,19 @@ impl IntoVCL<VCL_STRING> for &[u8] {
     fn into_vcl(self, ws: &mut Workspace) -> Result<VCL_STRING, VclError> {
         // Try to save some work if the buffer is already in the workspace.
         // We assume that &[u8] has always been readonly, so workspace data is valid.
-        if ws.contains(self) {
+        // However, we need to make sure that the buffer is null-terminated, either by its own last
+        // byte, or by the one after it
+        let with_extra_byte = unsafe { std::slice::from_raw_parts(self.as_ptr(), self.len() + 1) };
+        if (ws.contains(self) && *self.last().unwrap() == b'\0')
```
nyurik, line 273: rule of thumb - if you are doing `unwrap`, either it is wrong, or you need to have a comment explaining why it will never happen.  Especially when this is a result-returning function.  Empty slices are valid, so you may fail here because of that.

varnish/tests/vmod_test/src/lib.rs
```diff
@@ -140,6 +143,28 @@ mod rustest {
         }
     }
 
+    pub fn ws_tests(ctx: &mut Ctx) {
+        // external buffer -> new ptr
+        let buf = b"abc";
```
nyurik, line 148: add tests for an empty slice

gquintard: good callout

varnish-sys/src/vcl/convert.rs
gquintard: correct, I missed that one

gquintard: completely forgot that we have destructuring in `rust`, I love that thing

varnish-sys/src/vcl/convert.rs
```diff
@@ -267,13 +267,19 @@ impl IntoVCL<VCL_STRING> for &[u8] {
     fn into_vcl(self, ws: &mut Workspace) -> Result<VCL_STRING, VclError> {
         // Try to save some work if the buffer is already in the workspace.
         // We assume that &[u8] has always been readonly, so workspace data is valid.
-        if ws.contains(self) {
+        // However, we need to make sure that the buffer is null-terminated, either by its own last
```
nyurik, line 270: Could you explain why do we even support the `&[u8]`?  This actually seems like just something leftover from before that we should remove.  The `&[u8]` by definition has nothing to do with NULL-terminated strings.  The `CString/String` types on the other hand do imply that.

gquintard: mainly because the C definition of a string and the rust definitions are different, and also, `&[u8]` is what the workspace is. So for example, when we need to iterate of headers and stash them into the workspace, this comes in handy

nyurik: Let's keep concerns separate to make sure we have the best API. Just because workspace internally uses some byte representation should not be exposed to the Rust land.

* When using from Rust code, we should offer API that's hard to misuse.
  * Providing `&[u8] -> VCL_STRING` is too easy to misuse as you have found out, because you were assuming it will treat buffers (blobs) as strings implicitly
  * If a **Rust** user wants to send strings to Varnish, they should use `String/CString`, but **not** buffers
  * If a Rust user wants to send a binary blob, they should only have `&[u8] -> VCL_BLOB` path
* When dealing with header iteration, Rust API could offer two paths:
  * iterator returns strings, or maybe `&CStr` pointing to internals of the workspace
  * iterator returns some new entity struct/enum that hides implementation detals, and offers methods to get its content (e.g. as &CStr if it is that), and also offer a way to copy it to another workspace, e.g. `fn copy_to(&self, ws: &mut Workspace)`

gquintard: I mean, the true solution is for the `rust` layer to provide a workspace-backed `CString`, but I'm pretty sure we are not getting that.

My big use really is just that code: https://github.com/gquintard/vmod-reqwest/blob/main/src/lib.rs#L291-L314
I don't want to have to do a detour via the heap, and I don't think there's a better solution than letting me stuff the workspace with whatever bytes I feel are right (yes, I'm a C developer, I know, I know)

nyurik: it wouldn't be a workspaced-backed CString, but we can get a `&CStr` pointing to workspace - if we really needed that. The issue is that if a user function returns any kind of CStr, it would get copied (unless we do a quick pointer check that the ref points to the current workspace - might be ok actually)

I did some digging into the whole header business. I hate "fun" standards.

Basically, we have 2 ways of getting the `http::HeaderValue`'s content:
* non-validated - `as_ref() -> &[u8]` gets us that, and all bets are off because we have no idea what that data has, and it looks no different than any random byte array including potentially the invalid ones.
* validated - `to_str() -> Result<&str, ...>` - the value gets validated by the header lib, but not copied. Users are certain of its content. When copying it to the workspace, we need to do two validations: that its a valid &str (i.e. only has UTF8), and that its a valid header aftrewards.
  * to_str simply checks all bytes against `b >= 32 && b < 127 || b == b'\t'`

So there is a safe method and a fast method:
* safe would mean `&[u8] -> VCL_BLOB` and `&str -> VCL_STRING` conversions, with all the extra validations, but the blob is not something easy to use, so most cases one would need to do `to_str` and then copy to WS.
* A bit faster but much less stable method is to add `http` library as an optional dependency, and simply allow `HeaderValue -> VCL_STRING` converter. In order to reduce bloat and conflicts, we can even support all http versions - e.g. add `http0` and `http1` features that would require v0.*.* and v1.*.*, accept the headers, and do the conversion... not sure it will work, but this method does hold some promises...

gquintard: yess to all that, but it doesn't really address my core issue, which is that I need to fill the workspace by aggregating items into a single string/header, and the workspace is a `&mut [u8]`.

So #170 provides a way to go from `http:Header` to `VCL_STRING`, but in truth it really just adds an extra step since I need to start with a `&[u8]`. It's just adding an extra validation which is welcome, but not the core of my concern here

nyurik: but why do you start with a `&[u8]`?

nyurik: ahem, I'm probably missing something.  Workspace is essentially an allocation arena, with some user-facing API.  
Most users would interact with it by adding `VCL_STRING` and `VCL_BLOB` types to it - i.e. `into_vcl(value, ws) -> VCL_type`.  The `ws.reserve` seems to be more of an internal / unsafe part of this that shouldn't be used directly unless user is in "unsafe" mode.

nyurik: P.S. As such, the `into_vcl` should be "safe-only" - i.e. will do all the proper checks/validations/handholding, and should not have any surprises like "add a blob `&[u8]`, but get back a `VCL_STRING` with magically added null termination because we think &[u8] is really a string and does not contain NULL characters or non-ASCII characters between `0x0..31` except `\t`, etc etc etc"

gquintard: The workspace is a way to keep the memory in the same spot to improve cpu caching, it also guarantees that the area will live until the task is doesn't. It's a general-purpose storage for anything local to a task, so it makes a lot of sense to use it to build your payload before sending it back to Varnish.

> Most users would interact with it by adding VCL_STRING and VCL_BLOB types to it

you are prescribing, not describing. In the C world, vmod writers avoid `malloc`'ing as much as they can, and if they can use the workspace they will

gquintard: I know I shouldn't poke the bear, but: I don't think there should be different `write_*()` method, we just need to be able to write anything `AsRef<[u8]>`, the source doesn't matter, only that we are creating a valid `VCL_STRING`.

nyurik: :bear: 
should there be any multi-byte validations for VCL_STRING?  e.g. can I write `.write(b"\xF0\x9F"); .write("b"\x98\x8A");` -- which together for a :smile: [symbol](https://www.compart.com/en/unicode/U+1F60A)?

nyurik: We allow `write!(output, ...args...)` macro - which expands to `output.write_fmt(format_args!(...args...))` - in other words, our buffer object has to implement this method to support any string writing. Additionally, outputs (like String) tend to support `push_str(&str)` methods to copy content as is to the buffer without extra formatting steps. If you want to also support `push_bytes` - sure. This is the pattern common to many systems - so I would be hesitant to invent something significantly different, e.g. `push(foo.as_bytes())`, as we would still need to support the formatting case

gquintard: `VCL_STRING` and `VCL_BLOB` do not have multi-byte consideration, so while the question is a very interesting one in general, it doesn't apply here I believ

nyurik: Per chat, I reworked this API:  #171

gquintard:
ok to merge?

nyurik (approved):
looks good, i did a minor code cleanup

varnish-sys/src/vcl/convert.rs
```diff
@@ -267,13 +264,22 @@ impl IntoVCL<VCL_STRING> for &[u8] {
...
+        fn contains(ws: &Workspace, buf: &[u8]) -> bool {
+            ws.contains(buf) && buf.last() == Some(&b'\0')
+        }
+
+        if contains(ws, self)
+            || contains(ws, unsafe { from_raw_parts(self.as_ptr(), self.len() + 1) })
+        {
             Ok(VCL_STRING(self.as_ptr().cast::<c_char>()))
         } else {
             Ok(VCL_STRING(ws.copy_bytes_with_null(self)?.b))
```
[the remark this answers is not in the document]
gquintard: nope, but I will use the `.ends_with` suggestion

PR #169, opened 2025-02-12
rework docs generation a bit

gquintard:
remove the code snippet for constructors as it was a bit awkward to have non-compilable VCL code in the example.

I'm not exactly sure about the constructor titles, maybe `type Contructor` would be better, and avoid the immediate repetition


examples/vmod_be/README.md
```diff
@@ -19,18 +19,11 @@ import be;
 import be from "path/to/libbe.so";
 ```
 
-### Object `parrot`
+### Constructor (`parrot`) `parrot be.parrot(STRING to_repeat)`
```
nyurik, line 22: uhm... this is awful!  how many parrots can you have in one line? :)

> Constructor (parrot) parrot be.parrot(STRING to_repeat)

examples/vmod_be/README.md
```diff
@@ -19,18 +19,11 @@ import be;
...
 
-```vcl
-// Create a new instance of the object in your VCL init function
-sub vcl_init {
-    new new = parrot.new(STRING to_repeat);
-}
-```
-
-#### Method `BACKEND backend()`
+#### Method `BACKEND .backend()`
```
nyurik, line 29: I think having an example of how to create an object auto-generated for novice users as part of the docs is a good thing to have, wouldn't it?

varnish/snapshots/docs_types@docs.snap
```diff
@@ -1,5 +1,6 @@
 ---
 source: varnish-macros/src/tests.rs
+snapshot_kind: text
```
nyurik, line 3: please update your insta version

examples/vmod_be/README.md
gquintard: Which one would you prefer?

```
### Constructor (`parrot`) `parrot be.parrot(STRING to_repeat)`
### Constructor `parrot be.parrot(STRING to_repeat)`
### Constructor `be.parrot(STRING to_repeat)`
```

examples/vmod_be/README.md
gquintard: yes and no, I can definitely see people copy-and-pasting blindly and being shocked that the code doesn't compile

examples/vmod_be/README.md
nyurik: probably the last one/

gquintard: let's go with that then

examples/vmod_be/README.md
nyurik: I am sure we can come up with some clearer syntax like `/* STRING to_repeat, */` or something similar - but this is still much better starting point than doing it from scratch, i.e. figuring out that it has to be part of `vcl_init`

gquintard: I disagree, giving the wrong code is worse that giving no code. Two things we can do though:
- point at https://varnish-cache.org/docs/trunk/reference/vmod.html#objects-and-methods
- let the vmod writer provide examples
we don't have to reinvent the universe to make that pie, docs exist and it's ok to assume some familiarity with vmods

nyurik: I am not saying we should give **working** code. What I am proposing is to give something that looks like code - a path I have seen in many demos/documentations, and clearly state what has to be done to make the code usable, i.e. some commented out parameters with regular text

examples/vmod_object/README.md
```diff
@@ -19,19 +19,12 @@ import object;
...
-    new new = kv.new([INT cap]);
-}
-```
+### Constructor `object.kv([INT cap])`
 
 Create a new key-value store, with an optional capacity.
 If `cap` is 0 or less, it will be ignored.
 
-#### Method `STRING get(STRING key)`
+#### Method `STRING .get(STRING key)`
```
nyurik, line 27: that's not right...

gquintard:
@nyurik , I'm confused, CI is complaining about snapshots differring, but I'm running `just ci-test` and `just ci-test-extra` and they are both passing

examples/vmod_be/README.md
nyurik: oki, perhaps you are right - lets force the user to actually write proper docs anyway :)

examples/vmod_object/README.md
nyurik: * `<instance_name>.get(...)` ?
* `<variable>.get(...)`
* `<var>.get(...)`
* `<obj>.get(...)`

But i do see your point here

nyurik (approved):
i don't see any issues with the code, and the actual docs might be a bit cleaner, or not... hard to say what's best

examples/vmod_object/README.md
gquintard: I think I like `<obj>`, it's concise and it will catch your eye and force you to think. I'll go with that as a test

nyurik: perhaps `<object>` to be even more precises?

nyurik: (acronyms are evil in docs ;) )

gquintard: eh, you were the one offering!

nyurik:
> @nyurik , I'm confused, CI is complaining about snapshots differring, but I'm running `just ci-test` and `just ci-test-extra` and they are both passing

I just tried this branch - works ok for me. CI is also passing - so i guess it was resolved?

gquintard:
I had to use the `6.0` container, but yeah, eventually it worked. Sorry for the noise

examples/vmod_object/README.md
nyurik: I'm in an AI "reasoning" mode - first i spew out lots of raw ideas, and then go back and review them and give a better answer :-P

nyurik (approved):
looks good, thx!

PR #170, opened 2025-02-13
POC support for http::HeaderValue

nyurik:
See https://github.com/gquintard/varnish-rs/pull/168#discussion_r1953988116


nyurik:
I think http crate **is** the crate :)  Most of Rust projects I know use it.  Do note that this is an **optimization** only, possibly even opt-in.  Ignore the content of the proposed functions - I was simply showing the user-facing API here, not the content of these functions.

gquintard:
Honestly then, I'm not sure we need it then? The Varnish world uses its own http lingo. Unless you want to overhaul `varnish-rs` to use just `http`?

nyurik:
I think this PR might need to be significantly changed - instead switching to the [http crate](https://docs.rs/http/latest/http/#http-headers). It already has tons of time tested and fast logic to convert bytes into properly parsed name and value objects. The only downside, which might actually be insignificant in comparison, is that data will actually be copied to a heap-allocated storage.  Not ideal, but far saner than the current incorrect implementation.  This will also bring in all the standard HTTP header constants.

gquintard:
closing this for now, to be revisited if need be

PR #171, opened 2025-02-14
rework Workspace and Buffer APIs

nyurik:
- Introduce `Workspace::get_str_buffer() -> WsStrBuffer` instead of `reserve`
- There is no longer any need to write NUL bytes to the end of the buffer
- The new buffer implements the `Write` trait, and can be inspected for what has been written so far
- The buffer does not allow any access to "dirty" (unset) portion of the buffer
- The buffer must be finalized with `finish() -> VCL_STRING`

### DIFF
--- a/varnish-sys/src/vcl/ws.rs
+++ b/varnish-sys/src/vcl/ws.rs
@@ -27,8 +27,67 @@ use crate::ffi::WS_Inside;
 use crate::ffi::{txt, VCL_STRING};
 #[cfg(not(varnishsys_6))]
 use crate::ffi::{vrt_blob, WS_Allocated, VCL_BLOB};
-use crate::vcl::VclError;
+#[cfg(not(varnishsys_6))]
+pub use crate::vcl::ws_str_buffer::WsBlobBuffer;
+pub use crate::vcl::ws_str_buffer::{WsBuffer, WsStrBuffer, WsTempBuffer};
+use crate::vcl::{VclError, VclResult};
 use crate::{ffi, validate_ws};
+
+#[cfg(not(test))]
+impl ffi::ws {
+    pub(crate) unsafe fn alloc(&mut self, size: u32) -> *mut c_void {
+        assert!(size > 0);
+        ffi::WS_Alloc(self, size)
+    }
+    pub(crate) unsafe fn reserve_all(&mut self) -> u32 {
+        ffi::WS_ReserveAll(self)
+    }
+    pub(crate) unsafe fn release(&mut self, len: u32) {
+        ffi::WS_Release(self, len);
+    }
+}
+
+#[cfg(test)]
+impl ffi::ws {
+    const ALIGN: usize = align_of::<*const c_void>();
+    pub(crate) unsafe fn alloc(&mut self, size: u32) -> *mut c_void {
+        // `WS_Alloc` is a private part of `varnishd`, not the Varnish library,
+        // so it is only available if the output is a `cdylib`.
+        // When testing, VMOD is a lib or a bin,
+        // so we have to fake our own allocator.
+        let ws = validate_ws(self);
+        assert!(size > 0);
+        let aligned_sz = (size as usize).div_ceil(Self::ALIGN) * Self::ALIGN;
+        if ws.e.offset_from(ws.f) < aligned_sz as isize {
+            ptr::null_mut()
+        } else {
+            let p = ws.f.cast::<c_void>();
+            ws.f = ws.f.add(aligned_sz);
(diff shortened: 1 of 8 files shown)


gquintard:
Yay on the new interface, but as expected, nay on losing the ability to create a byte buffer (completely fine to rename the current interface though)

The `get_body` function you changed in `vmod_test| is an example why: a mod may have to store binary data and we don't want to impose string-like constraints on it.

nyurik:
I don't have any issues with storing a binary data - we will just have to have a different VCL_BLOB-generating finalizer.  In other words, we should have a clear separation of string-storage and blob storage.

nyurik:
I just added blob-based storage too - this way we can have totally different implementations for blob and str storage

gquintard:
Just &[u8] please, VCL_BLOB is just way for us to wrap it for Varnish if we ever deliver the data to it. There's no need to preemptively create a blob if we just use the data internally

varnish-sys/src/vcl/ws_str_buffer.rs
```diff
@@ -46,184 +141,252 @@ impl<T> WsBuffer<'_, T> {
...
+        Ok(())
+    }
+
+    pub fn extend_from_slice(&mut self, slice: &[Item]) -> VclResult<()> {
+        if self.buffer.len() < slice.len() {
+            return Err(WsOutOfMemory(NonZeroUsize::new(slice.len()).unwrap()));
+        }
+        unsafe {
+            let end = self.buffer.as_mut_ptr();
+            ptr::copy_nonoverlapping(slice.as_ptr(), end, slice.len());
```
gquintard, line 165: this could be a problem, at least we should document it

varnish-sys/src/vcl/ws_str_buffer.rs
```diff
@@ -46,184 +141,252 @@ impl<T> WsBuffer<'_, T> {
...
-        assert_eq!(get_cstr(&res), c"0123456789");
+        assert_eq!(buf.remaining(), 149);
+        // saving 10 bytes + nul
+        assert_eq!(get_cstr(&buf.finish()), c"0123456789");
 
-        // second reservation get 160 - 10 bytes
-        let mut buf = ws.get_str_buffer();
-        assert_eq!(buf.remaining(), 150);
+        let mut buf = ws.get_str_buffer().unwrap();
+        assert_eq!(buf.remaining(), 160 - round_up_to_8(10 + 1) - 1);
```
gquintard, line 290: I think you are rounding up because of the alignment, which depends on the CPU arch?

varnish-sys/src/vcl/ws.rs
```diff
@@ -192,24 +230,28 @@ impl<'a> Workspace<'a> {
...
+    pub fn get_blob_buffer(&mut self) -> VclResult<WsBlobBuffer<'a>> {
+        unsafe { WsBlobBuffer::new(validate_ws(self.raw)) }
+    }
+
+    /// Allocate workspace free memory as a temporary vector-like buffer
+    /// until [`WsTempBuffer::finish()`] is called.  The buffer is not intended
+    /// to be returned to Varnish, but may be shared among context users.
+    /// The buffer is returned as a `RefCell<&'a [T]>` to allow mutable access,
+    /// while tying the lifetime to the workspace.
+    pub fn get_temp_buffer<T>(&mut self) -> VclResult<WsTempBuffer<'a, T>> {
```
gquintard, line 253: that name is wrong, it's not more temporary than the other two

varnish-sys/src/vcl/ws_str_buffer.rs
```diff
@@ -0,0 +1,391 @@
...
+    ws: &'a mut ffi::ws,
+    /// The start of the writable buffer, aligned to the content type. Will set to null when finished.
+    start: *mut Item,
+    /// The reserved buffer will move its start as we write to it
+    buffer: &'a mut [Item],
+
+    _prefix: PhantomData<Prefix>,
+    _output: PhantomData<Output>,
+    _suffix: PhantomData<Suffix>,
+}
```
gquintard, line 28: it's quite a big dance just to allow `VCL_BLOB`, I'd much rather we had a `new_blob(&Ctx, &[u8])` constructor.

If you really have to, then you can add a `finalize_as_blob()` to the byte buffer, and stash `vrt_blob` at the end of the buffer, but this whole machinery makes me uneasy

gquintard (changes requested):
some changes discussed out-of-band are needed, but we are good

gquintard (approved):

PR #173, opened 2025-02-20
Fix Clippy 1.85 lints + minor fix

nyurik:
I modified the logic of the `wrap_finish` a bit - now it will properly handle if `htc` is null already, and won't crash.


gquintard (approved):
great cleanup

PR #179, opened 2025-03-02
Ensure git is clean after the build

nyurik:
The build may update README files, and we may forget to commit it


gquintard (approved):
Works for me, but maybe we should just auto-commit the MD changes?

nyurik:
I would feel a bit uneasy when PR auto-fixes itself beyond a simple "remove spaces + reformat" by precommit.  Our code gen could be buggy, and this is basically another test that we generated correct readme.

PR #181, opened 2025-03-19
Add Varnish 7.7 support

nyurik:


gquintard:
it's ok as it is, but we should leverage the new vmod fields: https://varnish-cache.org/docs/7.7/whats-new/changes-7.7.html#changes-for-developers-and-vmod-authors

possibly using `git` information and the crate version if that's possible?

nyurik:
a crate should not assume it is running inside a git repo - although we could try to do some magical heuristic like (1) figure out the root of the current binary being built, and (2) call some `git` command to get the current hash value. Neither is very stable and error prone, plus it might not be worth the time doing it. That said, we could add some optional attrs like this: `vmod(version="...", version_hash="...")`  -- this way the user may use it in any way they may want.  Automation is obviously better, but I will leave it up to other contributors to implement if it is useful to anyone

gquintard (approved):

PR #182, opened 2025-03-26
headers can be non-utf-8

gquintard:
defensively downgrade header values to &[u8]. We will need a cleaner interface at some point, but this will at least prevent silly panics

this will notably help save https://github.com/gquintard/vmod-reqwest/issues/16


nyurik (commented):
Thx for tackling this!  Would it make sense to go in a slightly different direction here, and simply return `None` if it cannot be parsed?

gquintard:
closing in favor of #183

PR #183, opened 2025-03-26
Handle raw byte headers

nyurik:
This is based on @gquintard 's work in #182 - but introduces a new `StrOrBytes` header value.

Note that the current code still appends `NUL` after the header if it does not contain it.

### DIFF
--- a/varnish-sys/src/txt.rs
+++ b/varnish-sys/src/txt.rs
@@ -3,6 +3,7 @@ use std::slice::from_raw_parts;
 use std::str::from_utf8;
 
 use crate::ffi::txt;
+use crate::vcl::StrOrBytes;
 
 impl txt {
     /// Internal helper to create a `txt` struct from a byte slice.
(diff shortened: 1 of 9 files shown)


varnish-sys/src/txt.rs
```diff
@@ -45,19 +46,28 @@ impl txt {
...
-    pub fn to_str<'a>(&self) -> Option<&'a str> {
-        self.to_slice().map(|s| from_utf8(s).unwrap())
+    pub fn to_str<'a>(&self) -> Option<StrOrBytes<'_>> {
+        self.to_slice().map(StrOrBytes::from)
     }
 
     /// Parse the `txt` struct as a header, returning a tuple with the key and value,
     /// trimming the value of leading whitespace.
-    pub fn parse_header<'a>(&self) -> Option<(&'a str, &'a str)> {
+    pub fn parse_header<'a>(&self) -> Option<(&'a str, StrOrBytes<'_>)> {
```
gquintard, line 57: Why aren't we using `'_` for the last item?

varnish-sys/src/vcl/str_or_bytes.rs
```diff
@@ -0,0 +1,34 @@
+use std::str::from_utf8;
+
+#[derive(Debug)]
+pub enum StrOrBytes<'a> {
+    Utf8(&'a str),
+    Bytes(&'a [u8]),
+}
```
gquintard, line 7: I'm wondering, should we have a `None` variant here? That would simplify a bunch of interactions I feel like

varnish-sys/src/vcl/str_or_bytes.rs
```diff
@@ -0,0 +1,34 @@
...
+        StrOrBytes::Utf8(value.as_str())
+    }
+}
+
+impl<'a> From<&'a [u8]> for StrOrBytes<'a> {
+    fn from(value: &'a [u8]) -> Self {
+        from_utf8(value).map_or_else(|_| StrOrBytes::Bytes(value), StrOrBytes::Utf8)
+    }
+}
+
```
gquintard, line 26: to your earlier point, maybe we can have an `.str_or_none()` method?

varnish-sys/src/vcl/str_or_bytes.rs
nyurik: Do we want to allow `set_header("header", None)` ?

gquintard: that would be an `Error` I'd say

nyurik: If so, it should not be part of the enum

gquintard: I don't fully agree but I can see the sense in that

gquintard (approved):

varnish-sys/src/vcl/str_or_bytes.rs
nyurik: basic idea with rust enums vs C enums - do not encode "error conditions" in it, like "uninitialized" or "error". Only encode what a user should be allowed to pass in. For everything else, there is an error code as part of the result.

PR #191, opened 2025-03-31
probes need to return a JSON object when json is requested

gquintard:


nyurik (approved):
no idea what this does, so totally up to you :)

varnish-sys/src/vcl/backend.rs
```diff
@@ -226,6 +223,9 @@ pub trait VclBackend<T: VclResponse> {
             vsb.write(&state).unwrap();
             vsb.write(&"]").unwrap();
         } else {
+            if detailed {
+                return;
+            }
             vsb.write(&"0/0\t").unwrap();
             vsb.write(&state).unwrap();
         }
```
nyurik, line 231:
```suggestion
        } else if !detailed {
            vsb.write(&"0/0\t").unwrap();
            vsb.write(&state).unwrap();
        }
```

nyurik (commented):
P.S. I did make a suggestion on optimizing the code flow

PR #192, opened 2025-04-01
Hack together a logo

nyurik:
The logo is based on varnish-cache + the commonly-used rust gear... plus a few SVG hacks and fixes.

We may want to spend a bit more time with it and make it more like https://github.com/maplibre/maplibre-rs#readme

See also https://github.com/rust-lang/rust-artwork/blob/master/logo/rust-logo-gear-only.svg


gquintard (approved):
very cute

PR #194, opened 2025-04-25
Minor dependency updates and cleanups

nyurik:
* bumped bindgen-helper

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,7 +21,6 @@ jobs:
           - setup: varnish76
           - setup: varnish75
           - setup: varnish60lts
-
     env:
       RUST_BACKTRACE: 1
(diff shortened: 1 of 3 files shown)


gquintard:
Ssl is just for some downstream dependencies? If so, I'd rather not add it, and even move request to rutls if we have to

nyurik:
ssl seems like a frequent-enough component for service (https) style ecosystem that many vmods may target - seems like a nice to have small addition to simplify most vmod development scenarios... i'm not hard-set on this, but seems to be a nice little thing to reduce development papercuts

gquintard:
Well, give me a list of three downstream projects that use it and it goes in. Otherwise I think it would be jumping the gund

nyurik:
oki, removed

gquintard (approved):

PR #197, opened 2025-05-13
add vmod metadata 2.0 support

### LINKED ISSUE #196: vmod metadata 2.0 incompatibility
With the release of varnish 7.7.1 and 7.6.3, the json format of the vmod metadata has changed from version 1.0 to 2.0, and is not backwards compatible.

See: https://github.com/varnishcache/varnish-cache/commit/69259780b790fe287c6dad79d412dd1ab1526ec2

This prevents any varnish-rs vmod from properly loading in those versions as varnish rejects them with the error "Syntax != 2.0".

I have created a naive fix for this: https://github.com/dward/varnish-rs/commit/024d00d3c1f636223fce75a993c12733f647f2ed

It's not great as it hardcodes versions that it changed in, but I did not necessarily see a way to detect the requirement otherwise. I can send this as a pull if you like.

dward:
This fixes #196 by adding version detection for the changeover to 2.0


varnish-macros/src/gen_func.rs
```diff
@@ -372,12 +372,23 @@ impl FuncProcessor {
             // This ensures the string is properly escaped and surrounded by quotes
             default = default.to_string().into();
         }
-        let mut json_arg: Vec<Value> = vec![
-            vcc_type.into(),
-            arg_name.into(),
-            default,
-            Value::Null, // spec param is not used at this point
-        ];
+
```
nyurik, line 375: I wonder if this would work?

```rust
let mut json_arg: Vec<Value> = vec![
            vcc_type.into(),
            #[cfg(varnishsys_vmod_meta_1_0)]
            arg_name.clone().into(),
            arg_name.into(),
            default,
            Value::Null, // spec param is not used at this point
        ];
```

dward: @nyurik - That works, though the logic needs to be reversed. I've updated the pull.

varnish-macros/src/generator.rs
```diff
@@ -115,7 +115,12 @@ impl Generator {
     }
 
     fn gen_json(&self) -> String {
-        let mut header: Vec<Value> = vec!["$VMOD".into(), "1.0".into()];
+        let mut version = "2.0";
```
nyurik, line 118: same thing here - we don't want to introduce a mutable variable just because we have conditional compilation.  A better pattern might simply be something like this (if it works):

```rust
let mut header: Vec<Value> = vec![
    "$VMOD".into(),
    #[cfg(varnishsys_vmod_meta_1_0)]
    "1.0".into(),
    #[cfg(not(varnishsys_vmod_meta_1_0))]
    "2.0".into(),
];
```

dward: Yes - that works. Updated!

nyurik: just for completness sake, this pattern also works:

```rust
let version = if cfg!(varnishsys_vmod_meta_1_0) { "1.0" } else { "2.0" };
```

nyurik:
the tests seem to be failing because they need to be re-approved with `just bless`, and this has to be done from inside the docker containers too, i.e. `just docker-run-77` (or whichever versions have changed)

nyurik:
P.S. do let me know if you run into any trouble with accepting test result changes

dward:
@nyurik - Tests are failing on varnish75 because it's somehow sharing the same snapshot path as v7.6. I'm not familiar with insta, if you would like to give me some guidance here 😄

nyurik:
bleh, i just did #198 to make    this PR a bit more reviewable. Once merged, i will merge it into this one, and will fix multi-versioned issue.

nyurik:
I'll need to hack on this a bit more tonight to make it pass the tests

nyurik (approved):

gquintard (approved):

PR #199, opened 2025-05-14
chore: minor nits

nyurik:


varnish-macros/build.rs
```diff
@@ -6,7 +6,7 @@ fn main() {
 
     let ver = std::env::var("DEP_VARNISHAPI_VERSION_NUMBER");
     let (major, minor, patch) = parse_version(&ver.expect("DEP_VARNISHAPI_VERSION_NUMBER not set"));
-    println!("cargo:rustc-env=VARNISHAPI_VERSION_NUMBER={major}.{minor}.{patch}");
+    println!("cargo::rustc-env=VARNISHAPI_VERSION_NUMBER={major}.{minor}.{patch}");
```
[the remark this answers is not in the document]
gquintard: bad AI?

varnish-sys/Cargo.toml
```diff
@@ -16,6 +16,7 @@ readme.workspace = true
 name = "varnish_sys"
 
 [package.metadata.libvarnishapi]
+# FIXME: which version should be set here, and what uses it?
```
gquintard, line 19: it can be used by downstream projects: https://github.com/varnish-rs/vmod-rers?tab=readme-ov-file#all-platforms

options:
- remove it, and scrub the downstream READMEs
- leave it here and replace the FIXME with a more relevant comment

varnish-sys/build.rs
```diff
@@ -19,6 +19,7 @@ impl VarnishInfo {
     fn parse(bindings: PathBuf, varnish_paths: Vec<PathBuf>, version: String) -> Self {
         let (major, minor) = parse_version(&version);
 
+        // >= 7.0 .. < 7.6
```
gquintard, line 22: this is not what the `if` statement says (version 8 is coming)

gquintard (changes requested):

varnish-macros/build.rs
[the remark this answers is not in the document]
nyurik: not bad, just outdated. Single colon was used before

varnish-sys/Cargo.toml
nyurik: this is not the way to do it. There is a proper link-based system.

Define the name using `links`:

https://github.com/varnish-rs/varnish-rs/blob/f870c7bebaebc9f5c5aa5ee4308dbb8f1df880ca/varnish-sys/Cargo.toml#L4

Set any values using `cargo::metadata` in the build script:

https://github.com/varnish-rs/varnish-rs/blob/f870c7bebaebc9f5c5aa5ee4308dbb8f1df880ca/varnish-sys/build.rs#L83

Using the env var in any build script downstream:

https://github.com/varnish-rs/varnish-rs/blob/f870c7bebaebc9f5c5aa5ee4308dbb8f1df880ca/varnish/build.rs#L5

varnish-sys/build.rs
nyurik: ???  the current `if` statement says exactly that, unless i can't read code (which is possible...)

gquintard: you are correct, **I** can't read code

varnish-sys/Cargo.toml
nyurik: sure, but we can do it afterwards, as this is all versioned - and that's what versions are for :)

gquintard: yes, but I'd rather not add a `FIXME` asking for information that you now have :-)

nyurik: of course - i will remove it :)

BTW, the only realistic way to get that value AFTER the binary is built is to create some temporary file in the output:

```rust
// build.rs
fn main() {
    let version = std::env::var("DEP_VARNISHAPI_VERSION_NUMBER").unwrap();
    
    let out_dir = std::env::var("OUT_DIR").unwrap();
    let dest_path = std::path::Path::new(&out_dir).join("../../../build_metadata.json");
    std::fs::write(
        dest_path,
        format!(r#"{{"varnish_version":"{version}"}}"#)
    ).unwrap();
}
```

gquintard (approved):

PR #202, opened 2025-06-02
[just][ci] choose varnish version to install

gquintard:


Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = ["varnish", "varnish-macros", "varnish-sys", "varnish/tests/vmod_test"
 #
 #
 # !STOP!  Update the version in the varnish dependencies below!!!
-version = "0.5.0"  ## <<<----- ! STOP !!!
+version = "0.5.1"  ## <<<----- ! STOP !!!
```
nyurik, line 12: i guess that `STOP` message is not visible enough :(

gquintard: not really, specially because I wasn't trying to release, but just to get the CI to pass, so I didn't stop to look around.

I would be easier if all the same versions were in the same spot, it'd make things a lot clearer and visible I think

nyurik:
This was merged as part of #203

PR #222, opened 2025-07-11
Ci fixes

gquintard:


gquintard:
sounds good to me, but now I'm confused, who installs `cargo`? is that `taiki-e/install-action@v2`?

.github/workflows/ci.yml
```diff
@@ -19,6 +19,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - &stable_rust
```
nyurik, line 22: nooooo!  no magical YAML please!  esp because github might not support some of it.  KISS!  Copy pasting a single line is OK, and makes it readable

nyurik:
in theory, if this action checks if the current stable is already installed, it skips the installation - even if the current installation has extra components like fmt and clippy.  In the future, when it runs and the stable is ahead of the currently installed one, it might just install the minimal profile without the extra tools - at which point we MIGHT get a breakage. But lets deal with that if we ever encounter it.

gquintard:
ah, gotcha, thanks

PR #230, opened 2025-09-17
bless new error format

gquintard:


gquintard:
@nyurik, I'm goign to need you here, my local `rust` disagrees with the CI's (even though I'm using `just`)

nyurik:
do you get the same output when you run `just env-info` locally?

nyurik:
why are you running nightly rust locally??

gquintard:
I might have switched to it because I needed a feature or something and never switched it back

PR #231, opened 2025-09-17
[sys] include string.h to get memmove/memset definitions

gquintard:


justfile
```diff
@@ -4,10 +4,10 @@ main_crate := 'varnish'
 features_flag := '--all-features'
 
 # which version of Varnish to install by default. Update the `supported_varnish_vers` variable below.
-default_varnish_ver := '7.7'
+default_varnish_ver := '8.0'
 
 # Make sure to update CI with the changes.  The versions with 'r' suffix are Varnish Plus versions - must have all 4 numbers.
-supported_varnish_vers := '7.7  7.7.0  7.6  7.5  6.0  6.0.14r3'
+supported_varnish_vers := '8.0 7.7  7.7.0  7.6  7.5  6.0  6.0.14r3'
```
nyurik, line 10:
```suggestion
supported_varnish_vers := '8.0  7.7  7.7.0  7.6  7.5  6.0  6.0.14r3'
```

varnish-sys/src/vcl/processor.rs
```diff
@@ -310,7 +318,7 @@ impl<'c, 'f> FetchFilters<'c, 'f> {
         if self.find_position::<T>().is_none() {
             let instance = Box::new(new_vfp::<T>());
             unsafe {
-                ffi::VRT_AddVFP(self.ctx, instance.as_ref());
+                ffi::VRT_AddFilter(self.ctx, instance.as_ref(), ptr::null());
```
nyurik, line 321: are you just changing these to the new function? what if its for before v8?

gquintard: The `Filter` variants have been introduced in 7.2, if memory serves, so this covers all the supported version, save for 6.0

nyurik: if it does not cover all versions (6 is a version), you should introduce a new flag for this and just do conditional call here

nyurik (approved):
looks good!

PR #235, opened 2025-10-28
Turn Rust panics into Varnish panics

dridi:
This sets the VAS panic hook from the event function, regardless of the actual event (meaning at least twice per VMOD life cycle).

It is safe to do this unilaterally because events all occur from the same cache-main thread, and even if that was not the case, the hook is only set up once.

The panic hook is global, so it may be overridden by a dependency, though I suspect having a random crate set a panic hook under the program's feet is probably frowned upon.

I suspect but haven't verified that two distinct Rust VMODs will not share this global hook as they will each have their own copy of the Rust std runtime code supporting panics and other language constructs.

The hook itself must not panic, so it is not desirable to turn Rust strings into C strings. If the Rust panic was caused by a failed allocation, chances are we won't be able to make null-terminated copies of the function name, file name and panic message.

This could be solved with a VAS_FailTxt() function in Varnish Cache, replacing `const char *` arguments with `txt` arguments instead.

An empty event function is injected when the VMOD does not declare one.

### DIFF
--- a/varnish-macros/src/gen_func.rs
+++ b/varnish-macros/src/gen_func.rs
@@ -7,7 +7,7 @@ use quote::{format_ident, quote};
 use serde_json::{json, Value};
 use syn::Type;
 
-use crate::model::FuncType::{Constructor, Destructor, Event, Function, Method};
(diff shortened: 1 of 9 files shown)


gquintard:
iceboxing it for lack of recent movement

PR #236, opened 2025-11-20
Add bitmap type to vsc

mnutt:
I'm not sure why I didn't add Bitmap when originally opening the VSC pull request, but here it is.

I am defaulting `#[bitmap]` type to use `bitmap` format; if that's an issue I can leave it defaulted to `integer` like the others.

### DIFF
--- a/varnish-macros/src/metrics.rs
+++ b/varnish-macros/src/metrics.rs
@@ -81,12 +81,11 @@ impl FromStr for Format {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "integer" => Ok(Format::Integer),
-            "bitmap" => Ok(Format::Bitmap),
             "duration" => Ok(Format::Duration),
             "bytes" => Ok(Format::Bytes),
             _ => Err(format!(
                 // TODO: can s come from an untrusted source? If so, we should not include it
-                "Invalid format value '{s}'. Must be one of: integer, bitmap, duration, bytes"
+                "Invalid format value '{s}'. Must be one of: integer, duration, bytes"
             )),
         }
     }


varnish/tests/pass/metrics.rs
```diff
@@ -19,4 +19,12 @@ pub struct VariousStats {
     /// A more detailed description than the above oneliner could go here.
     #[gauge(level = "debug", format = "bytes")]
     memory: AtomicU64,
+
+    /// Some arbitrary bitmap
+    #[bitmap]
+    flags: AtomicU64,
+
+    /// A bitmap with integer format override
+    #[bitmap(format = "integer")]
```
gquintard: I think the format should even be valid here?

varnish-macros/src/metrics.rs
```diff
@@ -198,16 +199,19 @@ fn generate_metrics(fields: &FieldList) -> ProcResult<BTreeMap<String, VscMetric
...
-            panic!("Field {name} must have either #[counter] or #[gauge] attribute")
+        let has_bitmap = has_attr(&field.attrs, "bitmap");
+        let metric_type = match (has_counter, has_gauge, has_bitmap) {
+            (true, false, false) => MetricType::Counter,
+            (false, true, false) => MetricType::Gauge,
+            (false, false, true) => MetricType::Bitmap,
+            (false, false, false) => {
+                // FIXME: use errors collection
+                panic!("Field {name} must have either #[counter], #[gauge], or #[bitmap] attribute")
+            }
```
gquintard: this feels a be silly, but I don't think there's a more elegant way

gquintard (changes requested):
this looks good to me (I'd love to get your advice @nyurik though). I just have a couple of comments and I will need some tests in `vmod-test` just to make sure we can access the data

varnish/tests/pass/metrics.rs
mnutt: What do you mean? It seems like it would be technically valid to have a bitmap with format integer, even if unlikely to be useful?

gquintard: it's technically valid in the sense that the JSON presents the two things independently, but a bitmap is a bitmap, the format is meaningless. Like, a metric or a gauge can represent an amount of bytes or a duration, but of the formats we have, a bitmap can only be a bitmap.

Plus, but I think if you start `varnishstat` in `curses` mode, you will see that if the format isn't `b`, you won't get the bitfield representation

gquintard: I think that would be perfect

gquintard: Make it sane, we can release 6.0.0, not big deal

gquintard (approved):
Looks great to me. I'll merge this week, unless @nyurik sees something scary

examples/vmod_object/src/lib.rs
```diff
@@ -44,7 +44,7 @@ mod object {
                 .get(key) // look for key
                 // If not found, create a new empty string (no memory allocation, so can use without lambda)
                 // If found, convert the &str to a String
-                .map_or(String::new(), |v| v.value().to_string())
+                .map_or(String::new(), |v| v.value().clone())
```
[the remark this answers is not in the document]
gquintard: ah, fair, nevermind ripping it out

examples/vmod_counters/src/lib.rs
```diff
@@ -81,6 +85,28 @@ mod stats {
                 std::sync::atomic::Ordering::Relaxed,
             );
         }
+
+        pub fn set_flag(&self, bit: i64) {
```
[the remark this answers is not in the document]
gquintard: ah, I think I follow where you are going.  I think it's fine here, but `get_flags` is a bit wonky indeed

PR #239, opened 2025-12-02
allow VCL_BLOB arguments as &[u8]

gquintard:


gquintard:
bump @nyurik

examples/vmod_blobs/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Overview
+
+Starting with this version of varnish-rs, you can accept BLOB data as input arguments to your VMOD functions. In Rust, BLOBs are represented as `&[u8]` slices, providing safe and efficient access to binary data.
+
+## Features Demonstrated
+
+This example VMOD provides several functions that work with BLOB data:
+
+- `blob_length(data: &[u8]) -> i64` - Returns the length of a BLOB
```
nyurik: is blob data always readonly?

examples/vmod_blobs/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// ```vcl
+    /// if (blob_example.is_empty(blob)) {
+    ///     # Handle empty blob
+    /// }
+    /// ```
+    pub fn is_empty(data: &[u8]) -> bool {
+        data.is_empty()
+    }
+
+    /// Computes a simple checksum (sum of all bytes modulo 256)
```
nyurik: sha256 is not exactly simple...

examples/vmod_blobs/README.md
gquintard: it's not (it's C underneath!), but here it's implemented as read-only

examples/vmod_blobs/src/lib.rs
gquintard:
```suggestion
    /// Computes the SHA256 checksum of the blob
```

examples/vmod_blobs/README.md
nyurik: do you want to mention that it's readonly for now in the docs somewhere?

gquintard: actually, we have this line in `convert.rs`:

```//! | `&[u8]` | <- | `VCL_BLOB` |```

that should be enough?

gquintard: hum, except that block doesn't show in the docs, I'll do that in an other PR

PR #243, opened 2026-01-16
Native backends and directors

perbu:
This PR adds native backend and director support to varnish-rs, enabling VMODs to create backends and directors that are managed by Varnish's runtime rather than requiring synthetic backend implementations in Rust.

#### Core Implementation

Native Backend Support (`varnish-sys/src/vcl/native_backend.rs`)

-   `NativeBackend`: Wrapper around Varnish's native HTTP backend (`VRT_new_backend`)
-   `NativeBackendConfig`: Builder API for configuring backends with timeouts, connection limits, PROXY protocol, and health probes
-   Delegates all HTTP networking, connection pooling, and health probing to Varnish's built-in implementation

Director Support (`varnish-sys/src/vcl/director.rs`)

-   `VclDirector` trait: Interface for implementing load balancing/routing logic
-   `Director<D>`: FFI wrapper that registers a Rust director implementation with Varnish (`VRT_AddDirector`)
-   Comprehensive lifecycle documentation covering VCL warm/cold states, reference counting, thread safety requirements, and common pitfalls

Backend Set Helper (`varnish-sys/src/vcl/backend_set.rs`)

-   Thread-safe collection for managing multiple backends with proper reference counting via `VRT_Assign_Backend`
-   Helper methods for health checking (`any_healthy`, `healthy_list`, `healthy_count`)
-   Selection algorithms: round-robin (`pick_by_index`), weighted random (`pick_by_weight`), priority fallback (`pick_first_healthy`)
-   Uses `RwLock` for concurrent read access

Endpoint Configuration (`varnish-sys/src/vcl/endpoint.rs`)

-   `Endpoint`: Network address specification supporting IPv4, IPv6, Unix domain sockets, and dual-stack configurations
-   Includes PROXY protocol preamble support

#### Macro Changes

Extended the VMOD macro system to support raw VCL types (`VCL_BACKEND`, `VCL_ACL`, etc.) as function parameters:

-   Added `ParamTy::VclType` variant for representing raw Varnish types
-   Implemented automatic null pointer checking for optional VCL types (since the blanket `impl<T> From<T> for Option<T>` cannot be overridden)
-   Updated type parsing to recognize `VCL_*` identifiers in both input parameters and return types
-   Changed `ParamTy` from `Copy` to `Clone` to accommodate string storage

#### Examples

`vmod_native_be`: Demonstrates native backend creation with configurable health probes. Shows how to create backends where Varnish handles all HTTP operations.

`vmod_directors_rs`: Implements round-robin and fallback directors in pure Rust using the `BackendSet` helper. Demonstrates proper backend reference counting, health checking, and VCL lifecycle management.

#### Test Coverage

Includes 12 VTC (Varnish Test Case) files covering:

-   Basic backend/director functionality
-   Health probe configuration and behavior
-   Round-robin distribution
-   Fallback/priority selection
-   VCL reload scenarios
-   Concurrent access
-   Dynamic backend addition/removal
-   Sick backend handling

#### Files Changed

-   4 new core modules in `varnish-sys/src/vcl/`
-   3 modified macro files for VCL type support
-   2 complete example VMODs with 12 test files
-   16 module exports added to `varnish-sys/src/vcl/mod.rs`

### DIFF
--- a/varnish-macros/src/parser_args.rs
+++ b/varnish-macros/src/parser_args.rs
@@ -337,7 +337,9 @@ impl ParamTy {
             // Check for VCL_* pattern (after existing matches)
             let ident_str = ident.to_string();
             if ident_str.starts_with("VCL_")
-                && ident_str.chars().all(|c| c.is_ascii_uppercase() || c == '_')
+                && ident_str
+                    .chars()
+                    .all(|c| c.is_ascii_uppercase() || c == '_')
             {
                 return Some(Self::VclType(ident_str));
             }
--- a/varnish-sys/src/vcl/backend_set.rs
+++ b/varnish-sys/src/vcl/backend_set.rs
@@ -150,10 +150,7 @@ impl BackendSet {
             if inner.backends[i].0 == be.0 {
                 // Decrement refcount by assigning null
                 unsafe {
-                    ffi::VRT_Assign_Backend(
-                        &mut inner.backends[i],
-                        VCL_BACKEND(ptr::null_mut()),
-                    );
+                    ffi::VRT_Assign_Backend(&mut inner.backends[i], VCL_BACKEND(ptr::null_mut()));
                 }
 
                 inner.total_weight -= inner.weights[i];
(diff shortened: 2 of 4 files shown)


gquintard:
I need to have a deeper look, but this PR scratches the wrong surface I think:
- for directors, it duplicates a ton of code that already exists to create backends, I think we can be smarter than than
- I'm not convinced of the usefulness of simply wrapping `vrt_backend()` in such a simple way
- the AI comments really sound like AI and might be plain wrong at times
- I don't think we need to own `BackendSet`, this feels like something the vmod writer should implement

there are merits to making directors more ergonomic, and I'll use this PR as a reference

PR #245, opened 2026-01-20
Allow VCL_BACKEND as param

mnutt:
This allows for passing a VCL_BACKEND to a vmod. I'm using it for writing new experimental director types in varnish-rs.


gquintard:
closing in favor of #246

PR #246, opened 2026-01-28
Introduce new Backend types

gquintard:
This introduces a couple of new traits and types:
- `Director` and `VclDirector` to implement directors
- `NativeBackendBuilder` to create IP/UDS backends
- `BakendRef` as an abstract handle over backend and directors


gquintard:
Ok, this is starting to look ok. There's still a ton of work to do on documentation, but I have an issue regarding new examples: they won't compile in 6.0 (nor do I want them to). @nyurik, do you know how we could easily exclude them from CI?

PR #260, opened 2026-03-24
Add local_socket support to the VCL context

perbu:
This change exports the VRT_r_local_socket function from the Varnish C headers and provides a safe wrapper via a new local_socket method on the Ctx struct.

The new method allows VMODs to retrieve the name of the listener socket that received the current request (e.g., "a0" or "http-80"), mapping to the VCL variable local.socket. The implementation handles conversion from C strings to Rust string slices and returns None if the name is empty or null, which typically occurs in backend contexts where session information is unavailable.


perbu:
(This is the last change I needed for ghost).

varnish-sys/src/vcl/ctx.rs
```diff
@@ -86,6 +86,22 @@ impl<'a> Ctx<'a> {
             }
         }
     }
+    /// Return the name of the listener socket that received the current request.
+    ///
+    /// This corresponds to the VCL variable `local.socket` and returns the `-a` socket
+    /// name (e.g., `"a0"`, `"http-80"`). Returns `None` in backend context where the
+    /// session isn't available, or if the name is empty.
+    pub fn local_socket(&self) -> Option<&'a str> {
+        let raw = unsafe { ffi::VRT_r_local_socket(self.raw) };
```
gquintard, line 101: annoyingly, this explodes if called from `vcl_init{}`. I'll see what we can do

gquintard:
added a few commits here: https://github.com/perbu/varnish-rs/pull/1

perbu:
This has everything I need and more. ctx.local_endpoint() might be useful.

gquintard:
uh, I'll need to look at why the docs are failing

gquintard:
closing in favor of #285 (rebased)

PR #262, opened 2026-03-25
[internal] cleanup version_after_v6 blocks

gquintard:
we can remove one level of nesting now that conditional support for older versions is gone


gquintard:
I'm bypassing the review check as it's a very minor change and I have changes on top of this that are more interesting coming soon

PR #277, opened 2026-05-15
CI: add new version links

gquintard:
run `fmt` and `clippy` while at it


gquintard:
I'm going to bypass the review check here, this should be non-controversial changes and my next PRs will fail if CI isn't up-to-date

PR #285, opened 2026-05-16
Add local_socket support to the VCL context

gquintard:
supersedes #260


varnish-sys/src/vcl/ctx.rs
```diff
@@ -86,6 +86,38 @@ impl<'a> Ctx<'a> {
...
+    /// This corresponds to the VCL variable `local.socket` and returns the `-a` socket
+    /// name (e.g., `"a0"`, `"http-80"`). Returns an `Err` in backend context where the
+    /// session isn't available, or if the name is non-UTF-8.
+    pub fn local_socket(&self) -> Result<&'a str, VclError> {
+        // we're breaking abstraction here, but the other ways are to just reimplement the
+        // whole logic in rust (which is admittedly short), or to let the user crash
+        if self.raw.req.is_null() && self.raw.bo.is_null() {
+            return Err("local.socket isn't available in this context".into());
+        }
+        let raw = unsafe { ffi::VRT_r_local_socket(self.raw) };
```
perbu, line 101: Is `VRT_r_local_socket` guaranteed to return a non-NULL `VCL_STRING` once the `req`/`bo` guard above has passed? If the C side can ever return NULL here (and the same for `VRT_r_local_endpoint` below), then `<&CStr>::from(raw)` is UB / panics. From a quick read of the Varnish source the getters seem to fall back to `""` rather than NULL, but it would be good to confirm that contract.

your call.

varnish-sys/src/vcl/ctx.rs
```diff
@@ -86,6 +86,38 @@ impl<'a> Ctx<'a> {
             }
         }
     }
+
+    /// Return the name of the listener socket that received the current request.
+    ///
+    /// This corresponds to the VCL variable `local.socket` and returns the `-a` socket
+    /// name (e.g., `"a0"`, `"http-80"`). Returns an `Err` in backend context where the
+    /// session isn't available, or if the name is non-UTF-8.
+    pub fn local_socket(&self) -> Result<&'a str, VclError> {
```
perbu, line 95: Worth being explicit in the docs about which VCL subs this is valid in. The `req.is_null() && bo.is_null()` guard correctly rejects `vcl_init`/`vcl_fini`, but a caller reading the docs alone can't tell that, they'll see "backend context" as the only excluded case. Suggest mentioning that the value is only meaningful in subs that have an associated session (client subs and `vcl_backend_*`).

perbu:
green

perbu (approved):

varnish-sys/src/vcl/ctx.rs
gquintard: we're guaranteed it's never NULL:
```
 #define LOC(var,fld)                                            \
  VCL_STRING                                                      \
  VRT_r_local_##var(VRT_CTX)                                      \
  {                                                               \
          struct sess *sp;                                        \
                                                                  \
          CHECK_OBJ_NOTNULL(ctx, VRT_CTX_MAGIC);                  \
          if (ctx->req) {                                         \
                  CHECK_OBJ(ctx->req, REQ_MAGIC);                 \
                  sp = ctx->req->sp;                              \
          } else {                                                \
                  CHECK_OBJ_NOTNULL(ctx->bo, BUSYOBJ_MAGIC);      \
                  sp = ctx->bo->sp;                               \
          }                                                       \
                                                                  \
          CHECK_OBJ_NOTNULL(sp->listen_sock, LISTEN_SOCK_MAGIC);  \
          AN(sp->listen_sock->fld);                               \                   <- this guy
          return (sp->listen_sock->fld);                          \
  }
```

PR #286, opened 2026-05-16
Support building VMODs on macOS (closes #160)

### LINKED ISSUE #160: Building vmods on macos
Building vmods on native macos (darwin, aarch64) works fine, with a few caveats:

* `crate-type = ["cdylib"]` produces a `.dylib` on macos, which just needs to be renamed to `.so` for varnish to use it
* the vmod depends on symbols not present in `libvarnishapi` (only in `varnishd`) which gives the linker problems. [0] This can be worked around either with [a library](https://github.com/jeff-davis/cdylib-plugin.rs) or by manually adding a `rustflags = ["-C", "link-arg=-Wl,-undefined,dynamic_lookup"]` directive to `.cargo/config.toml`.

Separately, I was initially thinking I would write unit tests for my vmod, but running the unit tests would require the symbols that are only present in `varnishd`, right? If this is correct I think I could work around it by structuring the vmod into two crates: one for the core with unit tests, and another for the varnish bindings. I don't think there is anything to be done here, I just wanted to confirm.

[0] Missing symbols:
```
dyld[24082]: symbol not found in flat namespace '_BS_CHUNKED'
```

perbu:
This PR promotes MacOS to a first class citizen. This will likely impact future development. OTOH, being more Unixy is likely a good thing and opens the door to support other OSes going forward. 

VMODs reference symbols (`VRT_*`, `VSL*`, `WS_*`, `BS_*`, ...) that live in `varnishd`, not `libvarnishapi`. Linux resolves these at dlopen time via its flat namespace; macOS's two-level namespace requires resolution at link time, so `cargo build` fails with `Undefined symbols for architecture arm64`.

Adding `-Wl,-undefined,dynamic_lookup` on macOS fixes it. Landing that surfaced two more macOS-specific issues that #160 didn't mention — both were hidden behind the link failure.

## Changes

  - `.cargo/config.toml` — sets the linker flag on `cfg(target_os = "macos")`.
  - `run_vtc_tests!` macro — was `env!("LD_LIBRARY_PATH")`, which fails to compile on macOS where Cargo sets `DYLD_FALLBACK_LIBRARY_PATH` instead. Now reads the platform-appropriate  var via a `cfg`-selected const.
  - `examples/vmod_error/tests/test01.vtc` — `echo -n` is non-portable; macOS `/bin/sh` writes the `-n` literally. Switched to `printf`.
  - CI — new `test-macos` job on `macos-latest`, installs Varnish via Homebrew. Bypasses `just ci-test` because the Justfile's install/version recipes are Debian-specific.
  - `CONTRIBUTING.md` — documents the linker flag for downstream VMOD authors, since it has to be emitted from the cdylib crate itself (can't propagate from `varnish-sys`, which is an rlib).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,10 +97,29 @@ jobs:
       - name: Ensure docs can be built at docs.rs without varnish dependencies
         run: just test-doc
 
+  # Smoke test on macOS. brew ships current Varnish (9.0.x) with arm64 bottles,
+  # so installation is fast. We bypass `just ci-test` because the Justfile's
+  # version-detection and Varnish install recipes are Debian-specific (dpkg/apt).
+  test-macos:
+    name: Test on macOS
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Install Stable Rust
+        uses: dtolnay/rust-toolchain@stable
+      - if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+        uses: Swatinem/rust-cache@v2
+      - name: Install Varnish
+        run: brew install varnish
+      - name: Build workspace
+        run: cargo build --workspace --all-features --all-targets
+      - name: Run tests
+        run: cargo test --workspace --all-features
+
   # This job checks if any of the previous jobs failed or were canceled.
   # This approach also allows some jobs to be skipped if they are not needed.
   ci-passed:
-    needs: [ test, test-varnish-ver, test-msrv, test-docs-rs ]
+    needs: [ test, test-varnish-ver, test-msrv, test-docs-rs, test-macos ]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,8 +98,9 @@ jobs:
         run: just test-doc
 
   # Smoke test on macOS. brew ships current Varnish (9.0.x) with arm64 bottles,
-  # so installation is fast. We bypass `just ci-test` because the Justfile's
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -97,10 +97,29 @@ jobs:
...
+        uses: dtolnay/rust-toolchain@stable
+      - if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+        uses: Swatinem/rust-cache@v2
+      - name: Install Varnish
+        run: brew install varnish
+      - name: Build workspace
+        run: cargo build --workspace --all-features --all-targets
+      - name: Run tests
+        run: cargo test --workspace --all-features
+
```
gquintard, line 121: can we use `just test` and `just build` here?

varnish/src/varnishtest.rs
```diff
@@ -8,6 +8,16 @@ use std::process::Command;
...
+/// Name of the env var Cargo populates with the dynamic-library search path for
+/// test binaries. macOS uses `DYLD_FALLBACK_LIBRARY_PATH`; other Unix-likes use
+/// `LD_LIBRARY_PATH`. (Windows is not a supported VMOD host.)
+#[doc(hidden)]
+pub const CARGO_DYLIB_PATH_ENV: &str = if cfg!(target_os = "macos") {
+    "DYLD_FALLBACK_LIBRARY_PATH"
+} else {
+    "LD_LIBRARY_PATH"
+};
+
```
gquintard: I'm not sure it makes sense to isolate this in `varnishtest.rs` as it's not specific to `varnishtest`, I think you could inline the logic in `run_vtc_tests`

Just my two cents though, I won't block the PR on that ground

gquintard (approved):

PR #290, opened 2026-05-19
vmod_example: add WsStrBuffer example

### LINKED ISSUE #142: formatting macros
it would be good for performance to be able to `concat!()` and `format!()` string directly into a workspace or to a header, the detour through a `String` and therefore the heap is annoying

gquintard:
fixes #142


gquintard:
should be non-controversial, merging once the CI passes

PR #292, opened 2026-05-20
fix(backend): deregister director on VCL_EVENT_DISCARD

perbu:
Custom backends/directors registered via `VRT_AddDirector` were only removed from `vdire->directors` in `Backend::drop` / `Director::drop`. When the per-VCL VMOD object that owns the Backend outlives the fini (e.g. parked behind an `Arc` in a shared pool), Drop does not run before `vcl_KillBackends`, and its trailing
`assert(VTAILQ_EMPTY(&vdire->directors))` at cache_vcl.c:704 panics the child.

Match the built-in `vbe_dir_event` behaviour: call `VRT_DelDirector` synchronously from the `VCL_EVENT_DISCARD` handler. A `deregistered` flag on the new `BackendShim<S>` wrapper guards against a second `VRT_DelDirector` from Drop. The shim is `#[repr(C)]` with the user type first so existing `priv_` casts in the other `wrap_*` callbacks remain valid.

Includes a `leaky_parrot` test object that wraps its Backend in `ManuallyDrop` to deterministically reproduce the assertion without the discard handler fix.

### DIFF
--- a/examples/vmod_be/README.md
+++ b/examples/vmod_be/README.md
@@ -27,3 +27,13 @@ it only needs two functions:
 - `backend()`, so that we can produce a C pointer for varnish to use
 
 #### Method `BACKEND <object>.backend()`
+
+### Constructor `be.leaky_parrot(STRING to_repeat)`
+
+`leaky_parrot` wraps its [`Backend`] in `ManuallyDrop` so the per-VCL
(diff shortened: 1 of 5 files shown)


perbu:
I'm gonna close this. Claude insisted this was the reason why a vmod failed but that wasn't true.

PR #293, opened 2026-05-22
Rework vmod object constructors

gquintard:
The biggest change is a constructor doesn't need to be named "new", instead, any function that returns `Self` is a constructor. This opens the door to multiple constructors producing the same object in a single vmod. (internally, this translates to multiple `C` function sets, but all pointing to the same `rust` fucntions, so the cost is low).

Thie PR also introduces a *breaking* change*: constructors in VCL bear the same name as the `rust` function, including `new()` which isn't special-case anymore to be named as the `impl` block it's in.

To alleviate this, if necessary, a new `vcl_rename()` macro is introduced to expose the function with a different VCL name.

`vmod_kv` benefits from this with the renaming from `kv` to `Kv`, negating the need for `allow(non_camel_case_types)`.

Edit: I've also push some clean ups in the example vmods while at it


gquintard:
gentle bump

gquintard:
@nyurik, @perbu: deadman switched activated, I'll merge tomorrow unless you ask me to delay

PR #297, opened 2026-06-07
Add new Acl type

thomasklinger1234:
# Summary

Adds support for `VCL_ACL` and matching ACLs against IP addresses. This is primarily a wrapper around `VRT_acl_match` and the introduction of a new `Acl` parameter type for VMODs. 

# Design

## Rust wrapper 

First, a new simple wrapper around `VCL_ACL` is introduced to hold a reference to the `ffi::VCL_ACL` (rust pseudocode). Here is a sketch of the interface

```rust
struct Acl {
  raw: ffi::VCL_ACL
}

impl Acl {
  // NOTE: match is a keyword 
  fn matches(self, Ctx, SockAddr /* aka VCL_IP */) -> bool {
    unsafe  { ffi::VRT_acl_match(ctx.raw,  self.raw, addr); }
  }
}
```

## VMOD interface

Then, the `Acl` type is added to `varnish-macros` and supported as a function parameter:

```rust
#[varnish::vmod(docs = "README.md")]
mod firewall {
  pub fn is_allowed(ctx: &mut Ctx, allowlist: Acl, addr: Option<SocketAddr>) -> bool {
    return allowlist.matches(ctx, addr)
  }
}
```

As an alternative, the ACL matching could be exposed on `Ctx` like this:

```rust
#[varnish::vmod(docs = "README.md")]
mod firewall {
  pub fn is_allowed(ctx: &mut Ctx, allowlist: Acl, addr: Option<SocketAddr>) -> bool {
    return ctx.acl_match(acl, addr)
  }
}
```

## Checklist 

(sourced from review)

- [x] Add `Acl::vcl_ptr()` instead of `return self.raw`
- [x] Implement `<VCL_ACL> for Option<Acl>` in `convert.rs`
- [x] Add `Acl::name(&self) -> &CStr` -> discontinued, since it requires vendoring vcc_interface.h
- [x] Add tests in `varnish/tests/pass/function.rs`
- [x] Add `Ctx::acl_match()` as the primary interface for users

### DIFF
--- a/varnish-sys/src/vcl/acl.rs
+++ b/varnish-sys/src/vcl/acl.rs
@@ -1,0 +1,35 @@
+use std::net::SocketAddr;
+
+use crate::ffi;
+
+#[derive(Debug, Clone)]
+pub struct Acl {
+    pub raw: ffi::VCL_ACL,
+}
+
+impl Acl {
+    /// Test if the given address matches the ACL. If `addr` is `None`, a negative match is returned.
+    pub fn matches(&self, ctx: &crate::vcl::Ctx, addr: Option<SocketAddr>) -> bool {
+        assert!(!self.raw.0.is_null());
+
+        match addr {
+            Some(ip) => unsafe {
+                let mut sa_buf = vec![0u8; ffi::vsa_suckaddr_len];
+                crate::vcl::convert::write_ip_to_buf(ip, &mut sa_buf);
+                ffi::VRT_acl_match(ctx.raw, self.raw, ffi::VCL_IP(sa_buf.as_ptr().cast())) == 1
+            },
+            None => false,
+        }
+    }
+
+    /// Retun the `C` pointer to the underlying ACL.
+    pub unsafe fn vcl_ptr(&self) -> ffi::VCL_ACL {
+        self.raw
+    }
+}
+
+impl From<ffi::VCL_ACL> for Acl {
+    fn from(p: ffi::VCL_ACL) -> Self {
+        Acl { raw: p }
+    }
+}
--- a/varnish-sys/src/vcl/acl.rs
+++ b/varnish-sys/src/vcl/acl.rs
@@ -1,4 +1,4 @@
-use std::net::SocketAddr;
+use std::{ffi::CStr, net::SocketAddr};
 
 use crate::ffi;
 
@@ -22,24 +22,19 @@ impl Acl {
         }
     }
 
+    /// Return the VCL name of the underlying ACL.
+    pub fn name(&self) -> &CStr {
+        assert!(!self.raw.0.is_null());
+
+        unsafe {
+            let acl = *self.raw.0;
+            assert_eq!(acl.magic, ffi::VRT_ACL_MAGIC);
+            CStr::from_ptr(acl.name)
+        }
+    }
+
     /// Retun the `C` pointer to the underlying ACL.
     pub unsafe fn vcl_ptr(&self) -> ffi::VCL_ACL {
         self.raw
-    }
-}
-
-impl From<ffi::VCL_ACL> for Acl {
-    fn from(value: ffi::VCL_ACL) -> Self {
-        Acl { raw: value }
-    }
-}
-
-impl From<ffi::VCL_ACL> for Option<Acl> {
-    fn from(value: ffi::VCL_ACL) -> Self {
-        if value.0.is_null() {
-            return None;
-        }
-
-        Some(Acl { raw: value })
     }
 }
--- a/varnish-sys/src/vcl/acl.rs
+++ b/varnish-sys/src/vcl/acl.rs
@@ -27,19 +27,3 @@ impl Acl {
         self.raw
     }
 }
-
-impl From<ffi::VCL_ACL> for Acl {
-    fn from(value: ffi::VCL_ACL) -> Self {
-        Acl { raw: value }
-    }
-}
-
-impl From<ffi::VCL_ACL> for Option<Acl> {
-    fn from(value: ffi::VCL_ACL) -> Self {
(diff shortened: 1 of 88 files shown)


thomasklinger1234:
I also have two additions already prepared (but not polished yet):

- `examples/vmod_acls` as an example similar to `vmod_timestamp`
- adding the `Acl` type to `varnish-macros` but would like to wait on a review for the `Acl` type first

varnish-sys/src/vcl/acl.rs
```diff
@@ -0,0 +1,35 @@
...
+                let mut sa_buf = vec![0u8; ffi::vsa_suckaddr_len];
+                crate::vcl::convert::write_ip_to_buf(ip, &mut sa_buf);
+                ffi::VRT_acl_match(ctx.raw, self.raw, ffi::VCL_IP(sa_buf.as_ptr().cast())) == 1
+            },
+            None => false,
+        }
+    }
+
+    /// Retun the `C` pointer to the underlying ACL.
+    pub unsafe fn vcl_ptr(&self) -> ffi::VCL_ACL {
```
thomasklinger1234, line 10: Todo: Check if this is needed.

varnish-sys/src/vcl/acl.rs
```diff
@@ -0,0 +1,35 @@
...
+use crate::ffi;
+
+#[derive(Debug, Clone)]
+pub struct Acl {
+    pub raw: ffi::VCL_ACL,
+}
+
+impl Acl {
+    /// Test if the given address matches the ACL. If `addr` is `None`, a negative match is returned.
+    pub fn matches(&self, ctx: &crate::vcl::Ctx, addr: Option<SocketAddr>) -> bool {
```
thomasklinger1234: Design question: To Option or not? 

VMODs parameters require Option<SocketAddr> so the match needs to be done somewhere, either in the VMOD implementation or here.

gquintard:
Thanks for this! A few thoughts

WE probably want `Acl::vcl_ptr()` as it's a bit cleaner than returning `self.raw`, and I have the feeling we'll want to make the `raw` field in most structs private some time in the future. And you've already made it `unsafe`, so that's perfect.

Could you implement `From<VCL_ACL> for Option<Acl>` (since `VCL_ACL` is nullable, which is weird, but that's an upstream thing) in `convert.rs` please?

The `Acl` type is correct as is, it's just a pointer. One thing you can add is `Acl::name(&self) -> &CStr`, but I can live without it for now.

Imma need some tests in `varnish/tests/pass/function.rs` for:
- `Option<Acl>` (VCL needs to provide a `VCL_ACL` which could be `null`) and `Option<Option<Acl>>` (`VCL` may provide the `VCL_ACL` which could be `null`) for input
- `Option<Acl>` for output

Also, having looked at the C code, I now see that `vrt_ctx` is needed for `ffi::VRT_acl_match`, which...urgh. Anyway, for consistency, I think having `Ctx::acl_match()` is more consistent with the rest of the cure code base.

varnish-sys/src/vcl/acl.rs
thomasklinger1234: no, but it is preferred

thomasklinger1234:
> The `Acl` type is correct as is, it's just a pointer. One thing you can add is `Acl::name(&self) -> &CStr`, but I can live without it for now.

There is an unfortunate finding though: `vrt_acl` and `VRT_ACL_MAGIC` are defined in `vcc_interface.h`. which is not exported directly to the FFI bindings. I could copy over their declaration to `varnish-sys/c_code/wrapper.h` which works but I would prefer not to. Probably an issue for the upstream project.

gquintard:
I don't think that's a big issue? We have `vrt_acl` declared as an opaque type via `vrt.h`, and even better, we have `VCL_ACL` that we can use. As we only want to call `VRT_acl_match`, that should be enough?

thomasklinger1234:
> I don't think that's a big issue? We have `vrt_acl` declared as an opaque type via `vrt.h`, and even better, we have `VCL_ACL` that we can use. As we only want to call `VRT_acl_match`, that should be enough?

Its not an issue until you want to implement `Acl::name()` which needs access to `raw.name`. The current bindings do not expose the field and the `VRT_ACL_MAGIC` constant. Maybe I miss something but cargo would not compile without the addition and vcc_interface is not part of libvarnishapi.

I have added an implementation but am not too confident that the addition to `wrapper.h` is really good.

varnish-sys/src/vcl/acl.rs
gquintard: no `Option<SocketAddr>`, it's the job of the caller to get us a proper `SocketAddr`

gquintard:
ah, right, then drop `Acl::name()`, I don't think it's that useful. If there's a user need later, we can consider "vendoring" the types in `wrapper.h` later on (I'll sync with upstream in the meantime so that hopefully don't need to do that)

thomasklinger1234:
@gquintard I have now integrated all of your feedback (tests, ACL match interface etc) and rebased the PR as well. 

The last missing piece is the integration test for `Option<Option<Acl>>` which I am struggling to get right: `Unsupported argument type`. Do you have any hints what I am missing?

gquintard:
I might have miscommunicated :-) the C ACL type isn't nullable (NULL isn't a valid ACL) so we do NOT want to support Option<Option<Acl>>.

thomasklinger1234:
@gquintard understood, thanks for clarification! all tests now pass and I think the functionality is complete. There are still two final points to get approval for:

- Are the tests in `varnish/tests/pass` enough? 
- Should the `match` logic still be located in `Acl::match` and `Ctx::acl_match` or should we relocate it to `Ctx` only?

PR #310, opened 2026-08-03
Add probe support to NativeBackendBuilder

### LINKED ISSUE #308: Setting probes with NativeBackendBuilder
# Context

I tried to use `NativeBackendBuilder` to create a backend but could not get a probe attached to it. From reading https://github.com/varnish-rs/varnish-rs/blob/754452516429103f701bbadf112120e1f5780831/varnish-sys/src/vcl/backend/backend_main.rs#L606 it seems that this is either unimplemented as its always set to null and there is no builder method exposed.

# Use case

What I want to do (pseudocode):

```rust
let be_builder = NativeBackendBuilder::new_ip(be_name, addr)
   .hosthdr(...)
   .probe(be_probe)
```

where `be_probe` might be supplied by the VMOD caller:

```vcl
probe p1 {
 ....
}

new magicbackend = magic.new(probe=p1);
```

---

Given this is just missing and would be useful and there is no logic error from my side, I would like to prepare a PR for that: 

- [x] Probe is a supported parameter
- [x] NativeBackendBuilder is already functional, so no extensive design or code required
- [x] Possible using C code (for example in `activedns` and `dynamic` VMODs)

thomasklinger1234:
# Summary

Fixes #308.

VMOD authors can now set probes for native backends created with NativeBackendBuilder:

```rust
let native_backend_probe = match probe {
   None => ffi::VCL_PROBE(null()),
   Some(p) => p.into_vcl(&mut ctx.ws).expect("workspace for probe"),
};

let native_backend = NativeBackendBuilder::new_ip(&name, sock_addr)
   .probe(native_backend_probe)
   .build(ctx)
   .ok()?;
```

(mostly taken from `examples/vmod_native_backend`). 

Rationale for VCL_PROBE: The rest of the parameters are low-level parameters (such as CStr), so this change follows that pattern as well. Also, there is no real way with the current setup to convert a higher level `Probe` to `VCL_PROBE` without a workspace (using `IntoVCL`).

---

# Open questions

- [x] What kind of tests should be added?
- [x] Is the interface actually good or is there a better choice? (Especially around `ffi::VCL_PROBE` conversion for VMOD authors) -> I would like to bring the higher level `Probe` types here instead of a raw `VCL_PROBE`

### DIFF
--- a/examples/vmod_native_backend/src/lib.rs
+++ b/examples/vmod_native_backend/src/lib.rs
@@ -1,11 +1,67 @@
+use varnish::vcl::BackendRef;
+
 varnish::run_vtc_tests!("tests/*.vtc");
+
+pub struct DynamicBackend {
+    /// A native backend created alongside this `DynamicBackend`.
+    backend: BackendRef,
+}
 
 #[varnish::vmod(docs = "README.md")]
 mod native_backend {
+    use super::DynamicBackend;
     use std::net::SocketAddr;
+    use std::ptr::null;
+    use varnish::ffi;
     use varnish::vcl::Ctx;
+    use varnish::vcl::IntoVCL;
     use varnish::vcl::NativeBackendBuilder;
+    use varnish::vcl::Probe;
     use varnish::vcl::{BackendRef, NativeBackend};
+
+    impl DynamicBackend {
+        /// Create a backend with an optional probe attached from a socket address.
+        ///
+        /// This function demonstrates creating native backends at initalization.
+        /// The backend is for the VCL lifetime and can be reused..
+        pub fn new(
+            ctx: &mut Ctx,
+            #[vcl_name] vcl_name: &str,
+            addr: &str,
+            probe: Option<Probe>,
+        ) -> Result<Self, &'static str> {
+            let Ok(sock_addr) = addr.parse() else {
+                return Err("failed to parse addr");
+            };
+
+            // Create backend name from address
+            let name = format!("{vcl_name}_{sock_addr}");
+            let Ok(name_cstr) = std::ffi::CString::new(name) else {
+                return Err("failed to create native name");
+            };
+
+            let native_backend_probe = match probe {
+                None => ffi::VCL_PROBE(null()),
+                Some(probe) => probe.into_vcl(&mut ctx.ws).expect("no workspace left"),
+            };
+
+            let Ok(native_backend) = NativeBackendBuilder::new_ip(&name_cstr, sock_addr)
+                .probe(&native_backend_probe)
+                .build(ctx)
+            else {
+                return Err("failed to build native backend");
+            };
+
+            let backend = native_backend.as_ref().clone();
+
+            Ok(Self { backend })
+        }
+
+        /// Retrieve the configured native backend.
+        pub fn backend(&self) -> BackendRef {
+            self.backend.clone()
+        }
+    }
 
     /// Create a dynamic backend from a socket address.
     ///
--- a/examples/vmod_native_backend/src/lib.rs
+++ b/examples/vmod_native_backend/src/lib.rs
@@ -1,10 +1,10 @@
-use varnish::vcl::BackendRef;
+use varnish::vcl::NativeBackend;
 
 varnish::run_vtc_tests!("tests/*.vtc");
 
 pub struct DynamicBackend {
     /// A native backend created alongside this `DynamicBackend`.
-    backend: BackendRef,
+    backend: NativeBackend,
 }
 
 #[varnish::vmod(docs = "README.md")]
@@ -52,16 +52,14 @@ mod native_backend {
                 return Err("failed to create native backend");
             };
 
-            let native_backend_ref = native_backend.as_ref();
-
             Ok(Self {
-                backend: native_backend_ref.clone(),
+                backend: native_backend,
             })
         }
 
         /// Retrieve the configured native backend.
         pub fn backend(&self) -> BackendRef {
-            self.backend.clone()
+            self.backend.as_ref().clone()
         }
     }
 
--- a/examples/vmod_native_backend/src/lib.rs
+++ b/examples/vmod_native_backend/src/lib.rs
@@ -49,12 +49,14 @@ mod native_backend {
                 .probe(&native_backend_probe)
                 .build(ctx)
             else {
-                return Err("failed to build native backend");
+                return Err("failed to create native backend");
             };
 
-            let backend = native_backend.as_ref().clone();
+            let native_backend_ref = native_backend.as_ref();
 
-            Ok(Self { backend })
+            Ok(Self {
+                backend: native_backend_ref.clone(),
+            })
         }
 
         /// Retrieve the configured native backend.
--- a/examples/vmod_native_backend/tests/test04.vtc
+++ b/examples/vmod_native_backend/tests/test04.vtc
@@ -1,0 +1,41 @@
+varnishtest "Test dynamic backend health probe (intentionally sick)"
+
+server s1 {
+    rxreq
+    txresp
+} -start
+
+varnish v1 -vcl {
+    import native_backend from "${vmod}";
+
+    backend default none;
+
+    probe p1 {
+        .initial = 0;
+        .threshold = 64;
+        .url = "/";
+        .expected_response = 599;
+    }
+
+    sub vcl_init {
+        new be1 = native_backend.new("${s1_addr}:${s1_port}", probe=p1);
+    }
+
+    sub vcl_recv {
+        set req.backend_hint = be1.backend();
+        return(pass);
+    }
+} -start
+
+
+client c1 {
+    txreq -url "/probe"
+    rxresp
+    expect resp.status == 503
+} -run
+
+logexpect l1 -v v1 -d 1 -g raw {
+    expect * 0 Backend_health "^be1_${s1_addr}:${s1_port} Still sick"
+} -start
+
+logexpect l1 -wait
--- a/examples/vmod_native_backend/README.md
+++ b/examples/vmod_native_backend/README.md
@@ -31,3 +31,16 @@ set req.backend_hint = native_backend.create("${server_addr}:${server_port}");
 
 * `[STRING addr]`:
 Socket address string (e.g., "127.0.0.1:8080")
+
+## Object `DynamicBackend`
+
+### Constructor `native_backend.new(STRING addr, [PROBE probe])`
+
+Create a backend with an optional probe attached from a socket address.
+
+This function demonstrates creating native backends at initalization.
+The backend is for the VCL lifetime and can be reused..
+
+### Method `BACKEND <object>.backend()`
+
+Retrieve the configured native backend.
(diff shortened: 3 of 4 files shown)


gquintard:
I'm liking it so far. Can you create a test for https://github.com/varnish-rs/varnish-rs/tree/main/examples/vmod_native_backend just to make sure we exercise the code? I can do it if time is lacking on your side.

I'm planning a release very soon, I'd love to have this in.

thomasklinger1234:
> I'm liking it so far. Can you create a test for https://github.com/varnish-rs/varnish-rs/tree/main/examples/vmod_native_backend just to make sure we exercise the code? I can do it if time is lacking on your side.
> 
> I'm planning a release very soon, I'd love to have this in.

sure, i had something already in place for development. i'll have to create another method on the vmod though because it does not make sense to add a probe to an ephermal backend with per-task lifetime. especially if VTC tests should be extended as well. unless you are fine with "just excercising" the code itself. 

i'll push it and we can do a second round of review.

gquintard:
you are correct, ephemereal backends wouldn't make sense, let's have another function.
You can also stash it in `vmod_test` if you feel like this would be polluting the example vmod

thomasklinger1234:
@gquintard I added a sample implementation with probes in the `native_backend` VMOD. Its now mixing objects and native backends though (from a focus perspective), don't know if thats the right way. 

If you have improvements in mind or want to change anything, feel free to do so (on this branch)

gquintard:
looking good, can we get a VTC too?

thomasklinger1234:
> looking good, can we get a VTC too?

working on it! seems like the current implementation somehow "loses" the backend as its not listed with `varnishadm backend.list` so I'll need to fix this as well.

**Edit: ** @gquintard  the VTC test is now ready but I'm not sure if the current test is really the VTC-way.

For clarification: 

When running more or less this VCL:

```vcl 4.1;

import native_backend from "/Users/thomas.klinger/GolandProjects/hub/varnish-rs/target/release/libvmod_native_backend.dylib";

    backend default none;

    probe p1 {
        .request =  "GET / HTTP/1.0"
                    "Host: www.example.com"
                    "Connection: close";
        .window = 8;
        .initial = 7;
        .threshold = 6;
        .expected_response = 200;
    }

    sub vcl_init {
        new be1 = native_backend.new("104.20.23.154:80", probe=p1);
    }

    sub vcl_recv {
        set req.http.host = "www.example.com";
        set req.backend_hint = be1.backend();

        return (pass);
    }
```

I can see that `Backend_health` events are emitted for `be1` with `Still healthy`. However, when performing a HTTP request against Varnish, it errors out and marks the backend as unhealthy which is a bit surprising to me. 

```
         0 CLI            - Rd ping
         0 CLI            - Wr 200 19 PONG 1786106374 1.0
         0 Backend_health - be1_104.20.23.154:80 Still healthy 4---X-RH 8 3 8 0.089430 0.075005 "HTTP/1.1 200 OK"
         0 CLI            - Rd ping
         0 CLI            - Wr 200 19 PONG 1786106377 1.0
     32769 Begin          c sess 0 HTTP/1
     32769 SessOpen       c ::1 59072 a0 ::1 22222 1786106375.539048 21
     32769 Link           c req 32770 rxreq
     32771 Begin          b bereq 32770 pass
     32771 VCL_use        b boot
     32771 Timestamp      b Start: 1786106375.539384 0.000000 0.000000
     32771 BereqMethod    b GET
     32771 BereqURL       b /
     32771 BereqProtocol  b HTTP/1.1
     32771 BereqHeader    b User-Agent: curl/8.7.1
     32771 BereqHeader    b Accept: */*
     32771 BereqHeader    b X-Forwarded-For: ::1
     32771 BereqHeader    b Via: 1.1 S0184253 (Varnish/9.0)
     32771 BereqHeader    b host: www.example.com
     32771 BereqHeader    b X-Varnish: 32771
     32771 VCL_call       b BACKEND_FETCH
     32771 VCL_return     b fetch
     32771 Timestamp      b Fetch: 1786106375.539405 0.000021 0.000021
     32771 FetchError     b backend be1_104.20.23.154:80: unhealthy
     32771 Timestamp      b Beresp: 1786106375.539415 0.000030 0.000009
     32771 Timestamp      b Error: 1786106375.539419 0.000035 0.000004
     32771 BerespProtocol b HTTP/1.1
     32771 BerespStatus   b 503
     32771 BerespReason   b Backend fetch failed
     32771 BerespHeader   b Date: Fri, 07 Aug 2026 12:39:35 GMT
     32771 BerespHeader   b Server: Varnish
     32771 VCL_call       b BACKEND_ERROR
     32771 BerespHeader   b Content-Type: text/html; charset=utf-8
     32771 BerespHeader   b Retry-After: 5
     32771 VCL_return     b deliver
     32771 Storage        b malloc Transient
     32771 Length         b 282
     32771 BereqAcct      b 0 0 0 0 0 0
     32771 End            b 
     32770 Begin          c req 32769 rxreq
     32770 Timestamp      c Start: 1786106375.539213 0.000000 0.000000
     32770 Timestamp      c Req: 1786106375.539213 0.000000 0.000000
     32770 VCL_use        c boot
     32770 ReqStart       c ::1 59072 a0
     32770 ReqMethod      c GET
     32770 ReqURL         c /
     32770 ReqProtocol    c HTTP/1.1
     32770 ReqHeader      c Host: localhost:22222
     32770 ReqHeader      c User-Agent: curl/8.7.1
     32770 ReqHeader      c Accept: */*
     32770 ReqHeader      c X-Forwarded-For: ::1
     32770 ReqHeader      c Via: 1.1 S0184253 (Varnish/9.0)
     32770 VCL_call       c RECV
     32770 ReqUnset       c Host: localhost:22222
     32770 ReqHeader      c host: www.example.com
     32770 VCL_return     c pass
     32770 VCL_call       c HASH
     32770 VCL_return     c lookup
     32770 VCL_call       c PASS
     32770 VCL_return     c fetch
     32770 Link           c bereq 32771 pass
     32770 Timestamp      c Fetch: 1786106375.539562 0.000349 0.000349
     32770 RespProtocol   c HTTP/1.1
     32770 RespStatus     c 503
     32770 RespReason     c Backend fetch failed
     32770 RespHeader     c Date: Fri, 07 Aug 2026 12:39:35 GMT
     32770 RespHeader     c Server: Varnish
     32770 RespHeader     c Content-Type: text/html; charset=utf-8
     32770 RespHeader     c Retry-After: 5
     32770 RespHeader     c X-Varnish: 32770
     32770 RespHeader     c Age: 0
     32770 RespHeader     c Via: 1.1 S0184253 (Varnish/9.0)
     32770 VCL_call       c DELIVER
     32770 VCL_return     c deliver
     32770 Timestamp      c Process: 1786106375.539576 0.000363 0.000014
     32770 Filters        c 
     32770 RespHeader     c Content-Length: 282
     32770 RespHeader     c Connection: keep-alive
     32770 Timestamp      c Resp: 1786106375.539719 0.000506 0.000143
     32770 ReqAcct        c 78 0 78 251 282 533
     32770 End            c 
     32769 SessClose      c REM_CLOSE 0.003
     32769 End            c 
```

...and I cannot list the backend with `varnishadm backend.list -j -p`. I think I am missing something here to "register" the backend with the CLI thread.

examples/vmod_native_backend/src/lib.rs
```diff
@@ -1,12 +1,70 @@
...
+            };
+
+            let native_backend_probe = match probe {
+                None => ffi::VCL_PROBE(null()),
+                Some(probe) => probe.into_vcl(&mut ctx.ws).expect("no workspace left"),
+            };
+
+            let Ok(native_backend) = NativeBackendBuilder::new_ip(&name_cstr, sock_addr)
+                .probe(&native_backend_probe)
+                .build(ctx)
```
thomasklinger1234, line 50: This is not enough to register the backend with the CLI thread (for `varnishadm backend.list`)? From reading the Varnish code itself, it seems that the native backends are only visible when registered using a director (and its "list" method) or by having them as plain VCL backends. 

Looking for clarification to test this hypothesis :)

examples/vmod_native_backend/tests/test04.vtc
```diff
@@ -0,0 +1,41 @@
+varnishtest "Test dynamic backend health probe (intentionally sick)"
```
thomasklinger1234, line 1: Kinda wonky. I searched in the Varnish source code for similar examples and collected the "best" of the VTC test snippets.

examples/vmod_native_backend/src/lib.rs
gquintard: uh, that's interesting. I would expected them to just be picked up. I'll look into it

examples/vmod_native_backend/tests/test04.vtc
gquintard: eh, that looks ok enough for me :-)

gquintard:
found the issue :-)

https://docs.rs/varnish/latest/varnish/vcl/struct.Backend.html:
> Once created, a [Backend](https://docs.rs/varnish/latest/varnish/vcl/struct.Backend.html)’s sole purpose is to exist as a C reference for the VCL. **As a result, you don’t want to drop it until after all the transfers are done.** The most common way is just to have the backend be part of a vmod object because the object won’t be dropped until the VCL is discarded and that can only happen once all the backend fetches are done.

you can't just keep the ref around, the object still needs to be around. It's a bit annoying, and I'm really trying to get upstream to get the refcounting right, but in the meantime, we need to be careful about references, the same way the C code is.

What happened is because the vmod dropped the backend, if "deleted" the backend, but Varnish kept it around until the references were gone, so the `unhealthy` message was misleading.

I've updated your branch: https://github.com/varnish-rs/varnish-rs/pull/310/changes/a6c80057979b2ed81b3b96b348b9d25158266aa3 would you mind testing that change and validate it does everything you've ever dreamed of? (or that at least it works? ;-) )

examples/vmod_native_backend/src/lib.rs
gquintard: turns out, it was enough, but you dropped `native_backend` so it deregistered the backend

thomasklinger1234:
> found the issue :-)
> 
> https://docs.rs/varnish/latest/varnish/vcl/struct.Backend.html:
> 
> > Once created, a [Backend](https://docs.rs/varnish/latest/varnish/vcl/struct.Backend.html)’s sole purpose is to exist as a C reference for the VCL. **As a result, you don’t want to drop it until after all the transfers are done.** The most common way is just to have the backend be part of a vmod object because the object won’t be dropped until the VCL is discarded and that can only happen once all the backend fetches are done.
> 
> you can't just keep the ref around, the object still needs to be around. It's a bit annoying, and I'm really trying to get upstream to get the refcounting right, but in the meantime, we need to be careful about references, the same way the C code is.
> 
> What happened is because the vmod dropped the backend, if "deleted" the backend, but Varnish kept it around until the references were gone, so the `unhealthy` message was misleading.

That makes sense, guess I just learned a new thing here 😄 

> I've updated your branch: [a6c8005](https://github.com/varnish-rs/varnish-rs/commit/a6c80057979b2ed81b3b96b348b9d25158266aa3) would you mind testing that change and validate it does everything you've ever dreamed of? (or that at least it works? ;-) )

Awesome, it works with my local `default.vcl` with a running `varnishd` as well. Thanks for looking into it!

